PR #64, opened 2026-03-17
[extensions] Fix user_id foreign key constraint blocking all inserts

### LINKED ISSUE #56: UserID is a foreign key constraint in the SQL in extensions/household-knowledge, but that table isn't created as part of the OB1 setup
### Which contribution?

extensions/household-knowledge

### What happened?

I followed the OB1 setup on Saturday and verified it is working. On Sunday, I started following the extension instructions. When I went to test storing a new fact with the household-knowledge extension, I got the error below (in Claude). I posted on Discord, and someone else said they had the same thing happen, so I know it wasn't just me. It seems like the extension instructions are out of sync with the setup instructions or there is something that got missed in the extension instructions. I could try to patch, but I'm concerned that I'd end up doing it in a different way than intended and then would need to redo it.

_Let me add that vendor for you._
_Loaded tools, used Household Knowledge integration_

Hmm, it looks like the user ID isn't set up yet. Let me try listing existing vendors first to find the right user ID.

_List vendorsList vendors_

Interesting, the list call works but the add fails. Let me try with a different default user ID format.

_Used Household Knowledge integration_

The Household Knowledge tool requires a valid user ID that's already registered in the system, and I'm hitting a foreign key constraint. It seems like your user account may not be set up in the Household Knowledge database yet.
Do you know if you've previously configured or used this Household Knowledge integration? There might be a setup step needed to create your user profile before vendors can be added. You could check the MCP server settings to see if there's a user ID or registration step required.

### What did you expect?

that this would work without issue

### Steps to reproduce

1. Set up using OB1 setup instructions
2. Follow the instructions to setup the household knowledge extension
3. try to add a vendor

### Environment

_No response_

matthallett1:
## Summary

Fixes #56 — reported by @kevingoldsmith. All 6 extensions fail on insert operations (add vendor, add item, etc.) due to a foreign key constraint against `auth.users`.

## Root Cause Analysis

The schemas define `user_id UUID REFERENCES auth.users(id) ON DELETE CASCADE NOT NULL`, but the MCP servers connect using the **service role key** (not an authenticated user session). This means:

1. `auth.uid()` is always `null` in this context
2. Claude/ChatGPT doesn't know the user's Supabase auth UUID
3. The AI guesses or fabricates a UUID for the `user_id` tool parameter
4. The FK constraint rejects the insert because the UUID doesn't exist in `auth.users`
5. **Read operations work fine** — they just return empty results for non-matching user_ids (which is why `list_vendors` worked but `add_vendor` failed in the original report)

This is a design-level bug, not a typo — the FK pattern doesn't work with the service role key approach that all OB1 MCP servers use.

### Schema files (6 files)
- Removed `REFERENCES auth.users(id) ON DELETE CASCADE` from all `user_id` columns
- Kept `user_id UUID NOT NULL` — it's still a logical grouping field, just not enforced against auth

### MCP servers (6 files)
- Added `DEFAULT_USER_ID` environment variable read at server startup
- Returns 500 with clear error message if not configured
- Removed `user_id` from all tool parameter schemas — the AI no longer needs to guess a UUID
- All insert/query operations use the env var instead

### Documentation (1 file)
- Updated `extensions/household-knowledge/README.md` with new Step 2: Generate Your User ID
- Added `DEFAULT_USER_ID` to credential tracker
- Updated troubleshooting section

### Files touched
| Extension | schema.sql | index.ts |
|-----------|-----------|---------|
| household-knowledge | ✅ | ✅ |
| home-maintenance | ✅ | ✅ |
| family-calendar | ✅ | ✅ |
| meal-planning | ✅ | ✅ |
| professional-crm | ✅ | ✅ |
| job-hunt | ✅ | ✅ |

## Testing Status

**This has NOT been tested against a live Supabase instance yet.** The changes are mechanically straightforward (remove FK, move user_id from tool params to env var), but given the scope (13 files, 6 extensions), we want validation from someone who has a working OB1 setup.

@kevingoldsmith — you originally reported this bug and have a working Supabase instance. Would you be willing to test the household-knowledge extension with this branch? The steps would be:

1. Re-run the updated `schema.sql` in your SQL editor (drop + recreate the tables, or just run the new version if starting fresh)
2. Set the env var: `supabase secrets set DEFAULT_USER_ID=$(uuidgen | tr '[:upper:]' '[:lower:]')`
3. Redeploy the edge function
4. Try `Add a vendor: Mike's Plumbing, phone 555-1234` again

If that works, the same pattern applies to the other 5 extensions.

## Test plan

- [ ] Verify household-knowledge `add_vendor` and `add_household_item` no longer hit FK constraint
- [ ] Verify `search_household_items` and `list_vendors` still return results
- [ ] Verify error message when `DEFAULT_USER_ID` is not set
- [ ] Spot-check one other extension (e.g., home-maintenance `add_maintenance_task`)
- [ ] Confirm no `REFERENCES auth.users` remain in any extension schema

### DIFF
--- a/extensions/family-calendar/index.ts
+++ b/extensions/family-calendar/index.ts
@@ -34,6 +34,11 @@ app.post("*", async (c) => {
     Deno.env.get("SUPABASE_SERVICE_ROLE_KEY")!,
   );
 
+  const userId = Deno.env.get("DEFAULT_USER_ID");
+  if (!userId) {
+    return c.json({ error: "DEFAULT_USER_ID not configured" }, 500);
+  }
+
   const server = new McpServer({ name: "family-calendar", version: "1.0.0" });
 
   // Tool: add_family_member
@@ -41,7 +46,6 @@ app.post("*", async (c) => {
     "add_family_member",
     "Add a person to your household roster",
     {
-      user_id: z.string().describe("User ID (UUID)"),
       name: z.string().describe("Person's name"),
       relationship: z.string().optional().describe("Relationship to you (e.g. 'self', 'spouse', 'child', 'parent')"),
       birth_date: z.string().optional().describe("Birth date (YYYY-MM-DD format)"),
@@ -51,7 +55,7 @@ app.post("*", async (c) => {
       const { data, error } = await supabase
         .from("family_members")
         .insert({
-          user_id: args.user_id,
+          user_id: userId,
           name: args.name,
           relationship: args.relationship,
           birth_date: args.birth_date,
@@ -78,7 +82,6 @@ app.post("*", async (c) => {
     "add_activity",
     "Schedule an activity or recurring event",
     {
-      user_id: z.string().describe("User ID (UUID)"),
       family_member_id: z.string().optional().describe("Family member ID (null for whole family)"),
       title: z.string().describe("Activity title"),
       activity_type: z.string().optional().describe("Type: 'sports', 'medical', 'school', 'social', etc."),
@@ -94,7 +97,7 @@ app.post("*", async (c) => {
       const { data, error } = await supabase
         .from("activities")
         .insert({
-          user_id: args.user_id,
+          user_id: userId,
           family_member_id: args.family_member_id || null,
           title: args.title,
           activity_type: args.activity_type,
@@ -127,7 +130,6 @@ app.post("*", async (c) => {
     "get_week_schedule",
     "Get all activities for a given week, grouped by day",
     {
-      user_id: z.string().describe("User ID (UUID)"),
       week_start: z.string().describe("Monday of the week (YYYY-MM-DD format)"),
       family_member_id: z.string().optional().describe("Optional: filter by family member"),
     },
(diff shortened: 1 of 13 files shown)


matthallett1:
## Type-check verification

All 6 extensions pass `deno check` with zero type errors:

- ✅ extensions/household-knowledge/index.ts
- ✅ extensions/home-maintenance/index.ts
- ✅ extensions/family-calendar/index.ts
- ✅ extensions/meal-planning/index.ts
- ✅ extensions/professional-crm/index.ts
- ✅ extensions/job-hunt/index.ts

This confirms the refactor (removing `user_id` from tool params, adding `DEFAULT_USER_ID` env var) compiles correctly. **Live testing against a Supabase instance is still needed** — @kevingoldsmith, your setup would be ideal for validating the household-knowledge extension end-to-end.

kevingoldsmith (approved):
I tested the household-knowledge extension and it works fine now! Will try the others but may not get them all done tonight.

matthallett1:
thanks Kevin!

PR #80, opened 2026-03-17
fix(server): URL-decode MCP key query param to preserve literal + characters

### LINKED ISSUE #70: [bug] API keys with special characters break URL query string auth
## Description

When the MCP access key generated via `openssl rand -hex 32` (or similar) contains characters like `+`, `/`, or `=`, the key breaks when passed as a URL query parameter (`?key=...`). In URL query strings, `+` is interpreted as a space, causing authentication to fail silently.

## Impact

- Users whose generated keys happen to contain `+` get mysterious auth failures
- cURL may work (properly encodes `+` as `%2B`) while browser/connector URLs fail
- Hard to debug because the key "looks right" but doesn't match after URL decoding

## Reproduction

1. Generate an access key that contains `+` (may take a few tries with `openssl rand -base64 32`)
2. Set it as the MCP access key in Supabase secrets
3. Try to connect via Claude Desktop or ChatGPT using `?key=YOUR_KEY` in the URL
4. Connection fails — the `+` becomes a space on the server side

## Discord Context

From **#help-old** (Mar 16):
> **thomas_smith:** "My generated key had a `+` in it, and in URL query strings + means 'space'. I was struggling to sort that since cURL properly encoded the + as %2B."
> 
> "Begs the question of how I ended up with that secret since the docs say use `openssl rand -hex 32`..."

thomas_smith offered to submit a PR: "Yeah I can even do a PR for the one line change I made to the MCP to handle it."

## Suggested Fixes

**Option A (prevent):** Ensure the documented key generation command only produces URL-safe characters. `openssl rand -hex 32` should be safe (hex only), but if any guide suggests `-base64`, that can produce `+`, `/`, `=`.

**Option B (handle):** URL-decode the key on the server side before comparing, so `%2B` and `+` both work.

**Option C (both):** Fix the docs AND add server-side URL decoding as a safety net.

cc: thomas_smith — your PR offer is welcome!

clawtom:
Fixes #70.

## Problem

`URLSearchParams.get("key")` treats `+` as a space (HTML form-encoding behavior). When a user's `MCP_ACCESS_KEY` contains a literal `+` and they pass it as a URL query parameter — e.g., `?key=abc+def` — the server receives `"abc def"` instead of `"abc+def"`, causing a silent auth failure.

This is a real-world failure mode when key generation uses base64 encoding (which can produce `+`, `/`, and `=`).

## Fix

Parse the raw query string with `decodeURIComponent` instead of `URLSearchParams.get()`. `decodeURIComponent` decodes percent-encoded sequences (`%2B` → `+`) but leaves literal `+` untouched, matching how the key was stored.

```ts
// Before:
const provided = c.req.header("x-brain-key") || new URL(c.req.url).searchParams.get("key");

// After:
const rawKeySegment = new URL(c.req.url).search.slice(1).split("&").find((p) => p.startsWith("key="));
const keyFromQuery = rawKeySegment ? decodeURIComponent(rawKeySegment.slice(4)) : null;
const provided = c.req.header("x-brain-key") || keyFromQuery;
```

## Behaviour after fix

| Scenario | Before | After |
|---|---|---|
| Key `abc+def` via `?key=abc+def` | `"abc def"` ❌ | `"abc+def"` ✅ |
| Key `abc+def` via `?key=abc%2Bdef` | `"abc+def"` ✅ | `"abc+def"` ✅ |
| Key passed via `x-brain-key` header | unaffected ✅ | unaffected ✅ |
| Hex-only key (no special chars) | works ✅ | works ✅ |

The `x-brain-key` header path is unaffected — headers are never form-encoded.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -355,8 +355,12 @@ server.registerTool(
 const app = new Hono();
 
 app.all("*", async (c) => {
-  // Accept access key via header OR URL query parameter
-  const provided = c.req.header("x-brain-key") || new URL(c.req.url).searchParams.get("key");
+  // Accept access key via header OR URL query parameter.
+  // Note: URLSearchParams treats '+' as a space, so we parse the raw query string
+  // and use decodeURIComponent (which preserves literal '+') instead.
+  const rawKeySegment = new URL(c.req.url).search.slice(1).split("&").find((p) => p.startsWith("key="));
+  const keyFromQuery = rawKeySegment ? decodeURIComponent(rawKeySegment.slice(4)) : null;
+  const provided = c.req.header("x-brain-key") || keyFromQuery;
   if (!provided || provided !== MCP_ACCESS_KEY) {
     return c.json({ error: "Invalid or missing access key" }, 401);
   }
@@ -367,3 +371,4 @@ app.all("*", async (c) => {
 });
 
 Deno.serve(app.fetch);
+


justfinethanku:
Closing — the guide already prevents this by generating hex-only keys via `openssl rand -hex 32`. No special characters in the key means no URL encoding issues.

PR #104, opened 2026-03-22
[integrations] Fix Slack Capture duplicate entries from webhook retries

demarant:
## Contribution Type

- [x] Integration (`/integrations`)

## What does this do?

Adds deduplication to the Slack Capture Edge Function. Slack retries webhook delivery if it doesn't get a response within 3 seconds, and the function (embedding + metadata extraction) typically takes 4-5 seconds, causing duplicate rows in the database. The fix checks if a row with the same `slack_ts` already exists before processing, and returns early if found. Also updates the troubleshooting section to reflect the fix.

## Requirements

None new. Uses the same services as the existing Slack Capture integration:
- Supabase (Edge Functions + database)
- OpenRouter API (for embeddings and metadata extraction)
- Slack workspace + app

## Checklist

- [x] My contribution has a `README.md` with prerequisites, step-by-step instructions, and expected outcome
- [x] My `metadata.json` has all required fields

### DIFF
--- a/integrations/slack-capture/README.md
+++ b/integrations/slack-capture/README.md
@@ -182,6 +182,14 @@ Deno.serve(async (req: Request): Promise<Response> => {
     const messageTs: string = event.ts;
     if (!messageText || messageText.trim() === "") return new Response("ok", { status: 200 });
 
+    // Deduplicate: Slack retries webhooks if response takes >3s, so check if we already captured this message
+    const { data: existing } = await supabase
+      .from("thoughts")
+      .select("id")
+      .contains("metadata", { slack_ts: messageTs })
+      .limit(1);
+    if (existing && existing.length > 0) return new Response("ok", { status: 200 });
+
     const [embedding, metadata] = await Promise.all([
       getEmbedding(messageText),
       extractMetadata(messageText),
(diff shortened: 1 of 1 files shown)


papaechosierra:
I solved this on my version with an await() for the embedding+metadata call, and just immediately reply to slack with 
ok:200. That fixes the no reply issue without needing to make further duplication checks.

justfinethanku (commented):
Thanks for this contribution, Antonio! This is a well-executed bug fix that addresses a real issue with webhook retries.

## What's Good

✅ **Solid technical solution** — The deduplication logic is correct: checking for existing rows with the same `slack_ts` before processing prevents duplicates caused by Slack's 3-second timeout retry behavior.

✅ **Proper placement** — The check happens early (right after validation, before the expensive embedding/metadata calls), which is efficient and prevents wasted API calls.

✅ **Documentation updated** — The troubleshooting section now accurately reflects that the issue is fixed rather than telling users to manually delete duplicates.

✅ **Clean code** — Uses `.contains()` on the metadata JSONB column appropriately, limits to 1 row for efficiency, and returns the same `200 ok` response that Slack expects.

✅ **PR format** — Title follows `[integrations]` convention, description is clear about what was changed and why.

✅ **No new dependencies** — Uses existing Supabase query patterns, no new services or tools required.

## Technical Review

The deduplication implementation:
```typescript
const { data: existing } = await supabase
  .from("thoughts")
  .select("id")
  .contains("metadata", { slack_ts: messageTs })
  .limit(1);
if (existing && existing.length > 0) return new Response("ok", { status: 200 });
```

This is correct and idiomatic for Supabase JSONB queries. The `.contains()` method checks if the `metadata` column contains the specified key-value pair, which is exactly what's needed here.

**One observation:** The query could be slightly more defensive by checking for `existing !== null` explicitly, but the current implementation (`existing && existing.length > 0`) works because Supabase returns `null` for errors and an empty array `[]` for no matches — and both correctly fail the condition.

## Metadata Check

The `metadata.json` hasn't changed, which is correct — this is a patch-level fix to existing functionality, not a version bump scenario (unless the maintainers want to bump to 1.0.1).

## Verdict: Ready to merge

This is a clean, targeted fix for a documented edge case. The code change is minimal, safe, and solves the problem without side effects. The troubleshooting documentation now reflects reality.

**Nice work on your first contribution to OB1!**

PR #117, opened 2026-03-25
[schemas] Reflections — structured reasoning traces linked to thoughts

alanshurafa:
## Summary

- Adds `reflections` table with uuid PK, FK to `thoughts(id)`, and 5 reflection types (decision, analysis, evaluation, planning, retrospective)
- `upsert_reflection` RPC for insert-or-update by ID with metadata merging
- `match_reflections` RPC for semantic similarity search over reflection embeddings (HNSW index)
- Service-role grants, RLS enabled, auto-updated `updated_at` trigger

## Motivation

AI agents and personal knowledge systems benefit from recording not just decisions but the reasoning behind them. Reflections capture trigger context, options considered, factors weighed, and conclusions — making past deliberation searchable and reusable.

## Test plan

- [ ] Run `migration.sql` in Supabase SQL Editor
- [ ] Verify `reflections` table exists via `information_schema.tables`
- [ ] Verify both RPCs exist via `information_schema.routines`
- [ ] Insert a test reflection via `upsert_reflection` and verify it returns a uuid
- [ ] Update an existing reflection by passing `p_id` and verify metadata merge
- [ ] Verify `match_reflections` returns results when embedding is provided

### DIFF
--- a/schemas/reflections/README.md
+++ b/schemas/reflections/README.md
@@ -1,0 +1,119 @@
+# Reflections
+
+> Structured reasoning traces linked to thoughts — capture deliberation processes with trigger context, options considered, factors weighed, and conclusions reached.
+
+## What It Does
+
+Adds a `reflections` table and two RPCs (`upsert_reflection`, `match_reflections`) that let you capture and search structured reasoning traces. Each reflection links to a thought and records why a decision was made, what options were considered, what factors influenced the outcome, and what conclusion was reached.
+
+This is useful for AI agents that need to recall past reasoning, or for personal knowledge systems where understanding *why* a decision was made is as important as the decision itself.
+
+## Prerequisites
+
+- Working Open Brain setup ([guide](../../docs/01-getting-started.md))
+- The `thoughts` table must already exist
+- pgvector extension enabled (included in the migration)
+
+## Schema Overview
+
+### `reflections` table
+
+| Column | Type | Description |
+|--------|------|-------------|
+| `id` | `uuid` (PK) | Auto-generated via `gen_random_uuid()` |
+| `thought_id` | `uuid` (FK) | References `thoughts(id)`, set null on delete |
+| `trigger_context` | `text` | What prompted this reflection |
+| `options` | `jsonb` | Options or paths that were considered |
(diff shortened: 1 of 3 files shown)


matthallett1 (changes requested):
Good schema design for reflections. Two issues from Codex review:

**P1: Partial updates silently erase data.** `upsert_reflection` with an existing `p_id` but without `p_options`/`p_factors`/`p_confidence` will overwrite stored values with the defaults (`[]`, `[]`, `NULL`). A "change only the conclusion" update silently erases the original options/factors/confidence. Fix: use `COALESCE(excluded.options, reflections.options)` pattern or similar.

**P2: Non-idempotent trigger.** Everything else uses `IF NOT EXISTS`/`CREATE OR REPLACE`, but `CREATE TRIGGER reflections_set_updated_at` fails on re-run. Wrap in a DO block or use `DROP TRIGGER IF EXISTS ... CREATE TRIGGER`.

alanshurafa:
Review feedback from @matthallett1 addressed in commit cd4b50c:

1. **P1 — Partial updates erasing data:** `options` and `factors` now use `CASE WHEN excluded.x = '[]'::jsonb THEN reflections.x ELSE excluded.x END` to preserve existing values when defaults are passed. `confidence` now uses `COALESCE(excluded.confidence, reflections.confidence)` so null defaults don't overwrite.
2. **P2 — Non-idempotent trigger:** Added `DROP TRIGGER IF EXISTS reflections_set_updated_at` before `CREATE TRIGGER`.

Ready for re-review.

alanshurafa:
Closing this PR as part of the OB1 Alpha Milestone consolidation. This feature is being rebuilt as one of 12 clean, gate-compliant PRs that together form the alpha upgrade path. The consolidated PRs will be submitted once verified on the fork. See the full plan for details.

alanshurafa:
Closing as part of the OB1 Alpha Milestone consolidation. This feature is being rebuilt as one of 12 clean, gate-compliant PRs that together form the alpha upgrade path. The consolidated PRs will be submitted once verified on the fork.

PR #122, opened 2026-03-25
[core] Make capture_thought compatible with content-fingerprint-dedup

matthallett1:
## Summary
- Updates the core MCP server (`server/index.ts`) to attempt using the `upsert_thought` RPC if the user has installed the `content-fingerprint-dedup` primitive.
- If the RPC doesn't exist (returns `PGRST202`), it gracefully falls back to the standard `insert`.
- This ensures that users who install the deduplication primitive don't break their core MCP server's `capture_thought` tool due to the new unique constraint.

## Test plan
- [x] Verified that `capture_thought` successfully calls `upsert_thought` when the primitive is installed.
- [x] Verified that the embedding is correctly updated on the returned row.
- [ ] (Needs verification) Ensure that the fallback works correctly on a fresh instance without the primitive installed.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,12 @@ supabase/
 
 # Recipe credentials & runtime state
 .env
+.env.local
+.env.*.local
 credentials.json
 token.json
 gmail-sync-log.json
 sync-log.json
+
+# Deno lock files
+deno.lock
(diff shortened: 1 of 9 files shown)


justfinethanku:
Thanks for the work here. I’m going to close this PR because it isn’t a fit for OB1 as a community PR.

The main change edits the core MCP server, which is explicitly out of scope for community contributions in this repo, and the PR also bundles unrelated support-file churn across multiple folders. That makes it the wrong contribution shape for OB1 even though the compatibility goal itself is understandable.

If we want this behavior upstream, it needs to come through a focused maintainer-led path instead.

PR #127, opened 2026-03-26
[Core] Add delete_thought tool and expose thought IDs in search/list

asherdiamant:
## What this does

Adds a `delete_thought` tool to the core MCP server and surfaces thought IDs in `search_thoughts` and `list_thoughts` output so thoughts can be referenced for deletion.

**Problem:** Without a delete tool, stale or incorrect thoughts can only be superseded by capturing new "correction" entries, leading to clutter over time. There's no way to clean up the knowledge base from within an AI session.

**Solution:** Three small changes to `server/index.ts`:

1. **`delete_thought` tool** — Accepts a thought UUID, confirms the thought exists, deletes it, and returns a preview of what was removed
2. **`search_thoughts`** — Now includes `ID: <uuid>` in each result
3. **`list_thoughts`** — Now includes `[<uuid>]` prefix in each result, and selects the `id` column

The `match_thoughts` RPC already returns the `id` column — this just surfaces it in the tool output.

## What it requires

No new services or dependencies. Uses the existing `thoughts` table and Supabase client. The `delete_thought` tool requires `service_role` permissions (already used by all other tools).

## Tested

Deployed and tested on a live Supabase instance with the standard Open Brain schema. Confirmed:
- Deleting by valid ID returns confirmation with content preview
- Deleting a non-existent ID returns an error message
- IDs appear correctly in search and list output
- Existing tools (capture, search, list, stats) are unaffected

### DIFF
--- a/integrations/kubernetes-deployment/index.ts
+++ b/integrations/kubernetes-deployment/index.ts
@@ -133,12 +133,13 @@ server.registerTool(
       const client = await pool.connect();
       try {
         const result = await client.queryObject<{
+          id: string;
           content: string;
           metadata: Record<string, unknown>;
           similarity: number;
           created_at: string;
         }>(
-          `SELECT content, metadata, created_at,
+          `SELECT id, content, metadata, created_at,
                   1 - (embedding <=> $1::vector) AS similarity
            FROM thoughts
            WHERE 1 - (embedding <=> $1::vector) >= $2
@@ -157,6 +158,7 @@ server.registerTool(
           const m = t.metadata || {};
           const parts = [
             `--- Result ${i + 1} (${(t.similarity * 100).toFixed(1)}% match) ---`,
+            `ID: ${t.id}`,
             `Captured: ${new Date(t.created_at).toLocaleDateString()}`,
             `Type: ${m.type || "unknown"}`,
           ];
@@ -235,11 +237,12 @@ server.registerTool(
       const client = await pool.connect();
       try {
         const result = await client.queryObject<{
+          id: string;
           content: string;
           metadata: Record<string, unknown>;
           created_at: string;
         }>(
-          `SELECT content, metadata, created_at
+          `SELECT id, content, metadata, created_at
            FROM thoughts
            ${whereClause}
            ORDER BY created_at DESC
@@ -254,7 +257,7 @@ server.registerTool(
         const results = result.rows.map((t, i) => {
           const m = t.metadata || {};
           const tags = Array.isArray(m.topics) ? (m.topics as string[]).join(", ") : "";
-          return `${i + 1}. [${new Date(t.created_at).toLocaleDateString()}] (${m.type || "??"}${tags ? " - " + tags : ""})\n   ${t.content}`;
+          return `${i + 1}. [${t.id}] [${new Date(t.created_at).toLocaleDateString()}] (${m.type || "??"}${tags ? " - " + tags : ""})\n   ${t.content}`;
         });
 
         return {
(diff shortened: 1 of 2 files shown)


matthallett1 (changes requested):
Great addition — `delete_thought` fills a real gap and exposing IDs in search/list is the right prerequisite. Two things to address:

**1. Scope the PR to what's advertised.** The `capture_thought` refactor (two-step upsert → single parallel insert) is a separate change that isn't mentioned in the PR title or description. If you want to ship that optimization, please either:
- Split it into its own PR with a clear description of what changed and why, OR
- Revert the capture_thought changes in this PR and only ship delete_thought + ID exposure

**2. Keep `integrations/kubernetes-deployment/index.ts` in sync.** The k8s deployment still exposes the old 4-tool interface without `delete_thought` and without IDs in search/list results. After this PR, users on the self-hosted PostgreSQL deployment can't use the new delete workflow at all. Please add the same changes to the k8s server, or at minimum document the divergence in the PR description.

Happy to merge once these are addressed!

asherdiamant:
Rebased onto main branch which advanced in the meantime

alanshurafa:
Both review items from @matthallett1 have been addressed in commit ed29eda:

1. **Scope:** The `capture_thought` refactor has been removed. The current diff only contains `delete_thought` + ID exposure in search/list — nothing out of scope.
2. **K8s sync:** `integrations/kubernetes-deployment/index.ts` now includes all three changes (delete_thought tool, IDs in search_thoughts, IDs in list_thoughts).

Ready for re-review.

asherdiamant:
Can anyone review this? @matthallett1? It seems stuck in the process. I just brought it up to date with the main branch so it's ready.

asherdiamant:
By the way, after updating the branch, I received this error:

[Invalid workflow file: .github/workflows/ob1-gate.yml#L1](https://github.com/NateBJones-Projects/OB1/actions/runs/24194935282/workflow)
(Line: 52, Col: 14): Exceeded max expression length 21000

justfinethanku:
Thanks for putting this together. I’m going to close this PR because it isn’t suitable for OB1 as a community PR.

This introduces a destructive core MCP tool by editing `server/index.ts`, and OB1 treats core MCP server changes as upstream infrastructure rather than mergeable community contribution surface area. The current PR also mixes core runtime changes with an integration file, so the boundary is off even before the feature risk is considered.

If this is something we want upstream, it needs a separate product/design decision first.

asherdiamant:
> Thanks for putting this together. I’m going to close this PR because it isn’t suitable for OB1 as a community PR.
> 
> This introduces a destructive core MCP tool by editing `server/index.ts`, and OB1 treats core MCP server changes as upstream infrastructure rather than mergeable community contribution surface area. The current PR also mixes core runtime changes with an integration file, so the boundary is off even before the feature risk is considered.
> 
> If this is something we want upstream, it needs a separate product/design decision first.

@justfinethanku 
I'm confused. 
Matt's initial response to the PR was:
"Great addition — delete_thought fills a real gap and exposing IDs in search/list is the right prerequisite. ...
Happy to merge once these are addressed!"
Which I did. The addition of the integration change was to keep everything consistent at his request. 

So I am not sure what changed.

How can we prevent chaff and noise accumulation in OB1 otherwise? Is there another solution? I came up with it due to necessity, I'm sure others need it too. I don't need this specific change, but the need for the tool is real.

justfinethanku:
I think my agent framed it poorly. Sorry about that. 

This should not have been closed, and when I get back to the office, I will re-open it. 

But I do think it's important to make destructive actions less accessible. Depreciating and versioning in a database should be the priority, not deletion.

asherdiamant:
@justfinethanku 
I understand the concern, and will add a solution when you reopen the PR. I suggest adding a table for deleted thoughts, where they can be recovered from, and it can be purged after a certain period, or manually, at the user's discretion. I prefer that to adding a flag into the main table, but that's another possible solution, since it keeps the live thoughts table focused and minimal. But that is another possible solution.

alanshurafa:
Agreed that the ability to delete thoughts is an absolute necessity over time especially as the brain grows. Also agree that adding a "recycle bin is a great idea". Its common to have a retention policy for projects like this.

PR #135, opened 2026-03-30
[recipes] Clean up Life Engine: add state table, harden permissions, fix skill divergence

air-captain:
## Summary

- **Add `life_engine_state` key-value table** for runtime state (cron job ID, sleep schedule) — replaces the `cron_state` briefing type workaround with a proper system state table
- **Simplify Dynamic Loop Timing** from 6 tiers to 4 (15m / 30m / 60m / one-shot) for cleaner reschedule logic
- **Remove duplicate embedded skill from README** — Step 5 now points to `life-engine-skill.md` as the single source of truth, eliminating a 130-line divergent copy
- **Harden permissions approach** — restructure Step 6 to lead with `settings.json` allowlist (scoped, persistent) as the recommended default; demote `--dangerously-skip-permissions` to testing-only; replace broad `mcp__supabase__*` wildcards with specific tool names
- **Fix skill-to-schema consistency** — add `user_responded` update logic, timezone note, remove phantom `suggestion_type` and `cron_state` references
- **Fix platform references** — consistently say "Telegram or Discord" where both apply

## Test plan

- [ ] Cross-reference all table/column/type references in `life-engine-skill.md` against `schema.sql` — all 9 QA checks pass
- [ ] Verify `schema.sql` creates 6 tables with no destructive SQL
- [ ] Verify README contains no embedded skill code (only pointer to `life-engine-skill.md`)
- [ ] Verify `settings.json` allowlist includes all tools Life Engine needs (`reply`, `react`, `edit_message`, `gcal_list_events`, `gcal_get_event`, `search_thoughts`, `list_thoughts`, `thought_stats`, `execute_sql`, `CronCreate`, `CronDelete`)
- [ ] Run `/life-engine` skill to confirm it executes a full cycle with the updated instructions

### DIFF
--- a/recipes/life-engine/README.md
+++ b/recipes/life-engine/README.md
@@ -526,7 +526,10 @@ Pre-approve only the specific tools Life Engine needs, persisted in your config 
       "mcp__open-brain__search_thoughts",
       "mcp__open-brain__list_thoughts",
       "mcp__open-brain__thought_stats",
+      "mcp__open-brain__capture_thought",
       "mcp__supabase__execute_sql",
+      "Bash(date *)",
+      "Bash(curl -s *api.open-meteo.com*)",
       "CronCreate",
       "CronDelete"
     ]
@@ -560,7 +563,9 @@ claude --channels plugin:telegram@claude-plugins-official \
     mcp__open-brain__search_thoughts \
     mcp__open-brain__list_thoughts \
     mcp__open-brain__thought_stats \
+    mcp__open-brain__capture_thought \
     mcp__supabase__execute_sql \
+    'Bash(date *)' 'Bash(curl -s *api.open-meteo.com*)' \
     CronCreate CronDelete"
 ```
 
--- a/recipes/life-engine/README.md
+++ b/recipes/life-engine/README.md
@@ -40,14 +40,14 @@ This guide contains everything Claude Code needs to set up your entire Life Engi
 5. Pause for you to complete Telegram pairing (requires your phone)
 6. Run a test cycle to confirm everything works
 
-After setup, exit and relaunch with your channel:
+After setup, exit and relaunch with your channel (permissions are handled by `settings.json` — see Step 6):
 
 ```bash
 # Telegram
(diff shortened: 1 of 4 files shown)


matthallett1:
@claude review this PR

matthallett1 (changes requested):
Good cleanup overall — the duplicate skill removal and permissions restructure are solid wins. However, Codex review (gpt-5.4) caught three issues, two of which are functional regressions:

**P1: Bash missing from the unattended allowlist.** The skill now uses `curl` for weather fetches, but the recommended `settings.json` allowlist doesn't include `Bash`. The first morning weather check will block on a permission prompt, breaking the unattended loop for users who follow the docs. Fix: add `Bash(curl:*)` to the allowlist example.

**P1: Discord reply events ignored.** The skill rule at line ~240 only fires for `<channel source="telegram">`. On a Discord setup, check-in replies and YES/NO approvals never match, so `user_responded` stays false and the self-improvement logic breaks. The README advertises Discord as first-class, but the skill doesn't support it.

**P2: No upgrade migration.** Changing `user_id` from `UUID` to `TEXT` inside `CREATE TABLE IF NOT EXISTS` doesn't update existing installs. Users who already have the tables need explicit `ALTER TABLE ... TYPE text` migration steps, or they'll get runtime errors when storing Telegram/Discord identifiers.

Happy to re-review once these are addressed!

alanshurafa:
Review feedback from @matthallett1 has been addressed across commits 6edb477 and 26910eb:

1. **P1 — Bash permissions:** Broadened to `Bash(*)` since scoped patterns were fragile. Documented in settings.json allowlist example.
2. **P1 — Discord support:** Added Discord reply event matching alongside Telegram.
3. **P2 — Migration:** Added upgrade migration steps for existing installs (UUID → TEXT column change).

Ready for re-review.

justfinethanku:
Thanks for the outstanding contribution. We took a final pass on the packaging, README clarity, and overall fit with OB1, and this is in good shape to land.

I’m merging this now so people can start using it. Really appreciate the work you put into this.

PR #148, opened 2026-04-01
[extensions] Fix professional-crm Accept handling for Streamable HTTP

Promithius-DR:
## Problem

The `POST /mcp` handler currently rewrites incoming `Accept` headers in a way that can still trigger reconnect-heavy behavior on stateless edge functions.

## Fix

Force JSON-only responses for this extension’s MCP endpoint by rewriting the request header to `Accept: application/json` before it reaches `StreamableHTTPTransport`.

This keeps the endpoint in plain request/response mode and avoids the idle reconnect loop.

## Impact

- Prevents the repeated reconnect behavior on stateless edge functions
- Keeps invocation volume under control
- Does not change the extension’s tool behavior

### DIFF
--- a/extensions/professional-crm/index.ts
+++ b/extensions/professional-crm/index.ts
@@ -19,6 +19,23 @@ const app = new Hono();
 
 // POST /mcp - Main MCP endpoint
 app.post("*", async (c) => {
+  // Force JSON-only responses. SSE causes a reconnect loop on stateless edge functions:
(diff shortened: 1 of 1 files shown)


justfinethanku:
Thanks for the fix here. We want to merge it as soon as these small cleanup items are addressed:

- retitle the PR to the required bracketed format, for example `[extensions] ...`
- align the PR description with the actual final code, or make the code match the stated intent if this is meant to be a pure removal

If you want, enable maintainer edits and we can patch this directly.

PR #164, opened 2026-04-08
Add optional local Ollama MCP setup and publish-safe docs

vystartasv:
This PR adds an optional local deployment path for Open Brain users who want to run embeddings and lightweight metadata extraction through local Ollama instead of a paid hosted model provider.

What's included:
- a local MCP server for capture/search against Supabase
- env-based local launcher templates, with secrets removed from the repo
- setup docs for humans
- notes for agent users
- SQL setup files aligned to the local `nomic-embed-text` path (`vector(768)`)
- safer metadata extraction fallback behavior

Why this may help:
- reduces dependency on paid inference providers for some users
- makes local/self-hosted setups easier to understand and reproduce
- keeps the existing Supabase storage model while allowing local inference
- improves publish safety by moving local credentials to env-based templates

Important note:
- this is an optional path, not a replacement for the existing hosted approach
- for users who want even more control, this could also be a stepping stone toward a fully local deployment path in the future

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -1,0 +1,11 @@
+# Copy this file to .env.local and fill in your own values.
+# Do not commit .env.local.
+
+SUPABASE_URL=https://your-project-ref.supabase.co
+SUPABASE_SERVICE_ROLE_KEY=your_service_role_key
+MCP_ACCESS_KEY=generate_a_long_random_access_key
+
+# Local Ollama defaults
+OLLAMA_BASE_URL=http://127.0.0.1:11434
+OLLAMA_EMBED_MODEL=nomic-embed-text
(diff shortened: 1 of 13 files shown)


justfinethanku:
Thanks for putting time into this. I’m going to close this PR because it isn’t a fit for OB1 in its current form.

The core blockers are architectural and packaging-related: OB1 requires remote MCP via Supabase Edge Functions, not local MCP server setups, and community contributions need to live in the standard category subfolder structure with the required packaging. This PR also has an internal vector-dimension mismatch, so it’s not just a docs-only divergence.

There’s useful thinking here, but it would need to be reimagined around OB1’s remote-MCP contribution model to be mergeable.

PR #133, opened 2026-03-29
[extensions] Add update_professional_contact tool to CRM extension

swhitt:
## Contribution Type

- [x] Repo improvement (docs, CI, templates)

## What does this do?

Adds the missing `update_professional_contact` tool to Extension 5 (Professional CRM), as proposed in #93 but never implemented. Accepts `contact_id` plus any combination of optional fields (`name`, `company`, `title`, `email`, `phone`, `linkedin_url`, `how_we_met`, `tags`, `notes`, `follow_up_date`). Only provided fields are updated. Also fixes `follow_up_date` being unwritable despite `get_follow_ups_due` querying it.

## Requirements

Nothing new.

## Checklist

- [x] I've read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] My contribution has a `README.md` with prerequisites, step-by-step instructions, and expected outcome
- [x] My `metadata.json` has all required fields
- [x] I tested this on my own Open Brain instance
- [x] No credentials, API keys, or secrets are included

PR #169, opened 2026-04-09
Added a new tool to update professional contacts with detailed fields

ben-kotvis:
Added a new tool to update professional contacts with detailed fields.

## Contribution Type

- [ ] Recipe (`/recipes`)
- [ ] Schema (`/schemas`)
- [ ] Dashboard (`/dashboards`)
- [ ] Integration (`/integrations`)
- [ ] Skill (`/skills`)
- [x] Repo improvement (docs, CI, templates)

## What does this do?

This simply adds support for updating a contact in the professional CRM.

## Requirements

Nothing new is needed

## Checklist

- [x] My contribution has a `README.md` with prerequisites, step-by-step instructions, and expected outcome
- [x] My `metadata.json` has all required fields
- [x] If my contribution depends on a skill or primitive, I declared it in metadata.json and linked it in the README

### DIFF
--- a/extensions/professional-crm/README.md
+++ b/extensions/professional-crm/README.md
@@ -155,12 +155,13 @@ If you build Extension 6, the `link_contact_to_professional_crm` tool works in r
 ## Available Tools
 
 1. **`add_professional_contact`** — Add a contact (name, company, title, email, phone, linkedin_url, how_we_met, tags, notes)
-2. **`search_contacts`** — Search by name, company, or tags with ILIKE
-3. **`log_interaction`** — Log a touchpoint (contact_id, interaction_type, summary, follow_up_needed, follow_up_notes). Auto-updates contact's last_contacted timestamp.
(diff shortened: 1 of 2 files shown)


justfinethanku:
Thanks for the contribution here. I’m not merging this as-is, but there’s a lot of potential if it gets tightened up.

What needs to change:
- scope the update by both `id` and `user_id` so this can’t mutate another user’s contact record
- make updates truly partial so omitted fields don’t get wiped out
- clean up the response/error copy so it consistently says `update` instead of `add`

If you want, enable maintainer edits and we can help patch the safe-update behavior directly.

justfinethanku:
Thanks for the contribution here. I’m closing this PR now because the feature has effectively been superseded by `#133`, which we patched and merged.

The useful part of the idea did make it into OB1, so this is not a rejection of the underlying contribution. At this point keeping both PRs open would just create confusion around which implementation is the live one.

PR #184, opened 2026-04-16
[recipes] Apple Notes import via macOS automation

DavidROliverBA:
## What this does
Imports Apple Notes directly into Open Brain using JXA (JavaScript for Automation) — no manual export step. Notes are converted from HTML to Markdown, chunked at heading boundaries, and ingested with embeddings and metadata including web URLs and note-to-note links.

## What it requires
- macOS with Notes.app (built-in)
- Python 3.10+
- OpenRouter API key
- Working Open Brain setup

## Tested on
- macOS 15 (Sequoia)
- Both iCloud and "On My Mac" accounts

I have tested this on my own Open Brain instance.


alanshurafa:
Thanks for the contribution, and welcome. The Apple Notes recipe itself is solid — JXA extraction with no manual export step, HTML-to-Markdown with heading-boundary chunking, a `notes_parser.py` with tests, and you have tested it against both iCloud and On-My-Mac accounts. macOS-only is fine for a recipe.

One scope thing before this is ready: the PR also includes two files under `docs/superpowers/plans/` and `docs/superpowers/specs/` — those look like planning and design artifacts from your own build workflow, not project documentation (OB1's `docs/` is for setup guides, FAQ, and companion prompts). Recommend author refresh to drop those two files and keep the PR scoped to `recipes/apple-notes-import/`. The recipe is the contribution; the process docs do not need to ship with it.

PR #193, opened 2026-04-18
[recipes] Brain backup and export

alanshurafa:
## Summary

Node.js recipe that exports every Open Brain Supabase table to dated local JSON files with streaming pagination.

- Paginates through PostgREST 1,000 rows at a time; streams rows to disk so large tables never load fully into memory
- Dated filenames (`<table>-YYYY-MM-DD.json`) for simple daily retention
- Prints a row-count and file-size summary per table
- **`thoughts` is required; optional companion tables (`entities`, `edges`, `thought_entities`, `ingestion_jobs`, `ingestion_items`) are backed up only if present.** Stock Open Brain installs skip them cleanly instead of erroring.
- No dependencies beyond Node 18+; config via `.env.local`

## Why

Open Brain users accumulate thousands of thoughts over time but Supabase-side backups require plan-tier access or manual dumps. A zero-dependency Node recipe gives anyone on the free tier a scriptable backup path that runs from cron / Task Scheduler.

The optional-table handling lands specifically so this recipe works against **every** Open Brain install, not just ones that have also installed the entity-extraction or smart-ingest schemas. Missing companion tables return a PostgREST 404, which the script treats as "skip with a log line" rather than an error.

Part 11 of 12 in the OB1 Alpha Milestone consolidation.

## Test plan

- [ ] Run against a stock Open Brain project (only `thoughts` table) — verify `thoughts` backs up, optional tables log `skipped (table not present)`
- [ ] Run against a project that has the optional tables — verify all exports succeed
- [ ] Drop `thoughts` temporarily → verify the script throws `Required table "thoughts" not found`
- [ ] Kill mid-run on a large `thoughts` export → verify no partial/corrupt JSON left behind (next run overwrites cleanly)
- [ ] Verify dated filename rotates correctly across days
- [ ] `node --check backup-brain.mjs` passes
- [ ] `metadata.json` passes the gate schema


alanshurafa:
Conflicts with `main`. Before I rebase, should this stay as-is, be reshaped, or be dropped? I can refresh it quickly if you want to keep it.

alanshurafa:
Rebased onto `main` — no conflicts now. This updates the existing `recipes/brain-backup/` rather than adding it.

For review context: `main`'s copy of this recipe predates its code review. This PR is the reviewed version — terminating pagination (the loop on `main` checks a stale page-size value), PGRST205-only "table missing" handling, atomic writes via tmp+rename, an AbortController fetch timeout, and BOM-tolerant env parsing. An upgrade over what's on `main`.

justfinethanku (approved):
Reviewed the recipe/code updates, added the missing community credit block, and confirmed the refreshed branch passes the visible required Actions checks. Approving for maintainer merge.

PR #195, opened 2026-04-18
[skills] Auto-capture Claude Code adapter

alanshurafa:
## Summary

Concrete Claude Code adapter for the [\`skills/auto-capture/\`](../auto-capture/) behavioral protocol by @jaredirish. Where the base skill defines a verbal trigger ("wrap up", "park this") and the user runs \`capture_thought\` manually, this adapter fires the same capture automatically via Claude Code's Stop hook at session end.

- Zero-dependency Node 18+ ESM script
- Parses the Claude Code transcript file on session-end
- Wraps user content in \`<thought_content>\` delimiters before posting (prompt-injection defense)
- AbortController on every fetch (10s default, \`FETCH_TIMEOUT_MS\` env override)
- Retry queue for transient failures; 4xx permanent failures dumped to \`dead/\` on first try
- Idempotent via \`import_key\` (session-id + content hash) so retries don't duplicate

## Why

Jared's protocol is the right behavioral frame for "end of session, capture the gold." What it needs is a binding for the client that can fire the protocol without the user having to remember. Claude Code's Stop hook is that binding for anyone running Claude Code as their primary client. Other clients (Cursor, generic CLIs) can follow the same shape.

Install both skills together — the base skill teaches the protocol; this adapter mechanizes it.

## Test plan

- [ ] Install alongside \`skills/auto-capture/\`
- [ ] Configure Claude Code Stop hook per README Step 3
- [ ] Have a session with 3+ user turns, close it — verify a thought appears in the brain with \`source_type = "claude_code_session"\`
- [ ] Short session (< 3 user turns) — verify \`skipped:too_short\` appears in \`logs/ambient-capture.log\`, no thought captured
- [ ] Kill SUPABASE_URL mid-session-end — verify the capture lands in \`data/capture-retry-queue/\`, gets picked up on next session
- [ ] Invalid MCP_ACCESS_KEY — verify 401 produces \`error:http_401:permanent\` log (no infinite retry)
- [ ] \`node --check session-end-capture.mjs\`
- [ ] Verify \`metadata.json\` passes the gate schema


alanshurafa:
Mergeable, no conflicts against `main`. No blockers from my side; ready whenever it reaches the queue.

justfinethanku (approved):
Maintainer pass complete. Added the OB1 community badge + creator credit to the new adapter README, refreshed the branch against current `main`, and verified the refreshed head passed the repo label and markdown lint checks.

PR #98, opened 2026-03-22
[schemas] Ingestion jobs schema for document ingest

alanshurafa:
## Summary
- Adds `ingestion_jobs` and `ingestion_items` tables for tracking document ingestion lifecycle
- Includes `append_thought_evidence` RPC for idempotent evidence appending
- Supports dry-run workflow: extract and reconcile without writing, then execute approved items
- All IDs are UUID, RLS enabled (service-role only)

## Why
Document ingestion (extracting atomic thoughts from raw text) needs a tracking layer for the extract → deduplicate → reconcile → execute pipeline. This schema is the backbone that Smart Ingest and future extraction tools build on.

## Test plan
- [ ] Apply migration on a fresh Open Brain instance
- [ ] Create sample job + items
- [ ] Verify `append_thought_evidence` is idempotent on repeat calls
- [ ] Verify UUID flow works end to end

Tested against a production Open Brain instance with 75K+ thoughts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #196, opened 2026-04-18
[schemas] Smart ingest pipeline tables

alanshurafa:
## Summary

Companion schema for the [smart-ingest integration] (separate PR) — two tables that back async document-ingestion with dry-run/execute semantics:

- **\`ingestion_jobs\`** — one row per ingest request, tracks status, fingerprint, counts
- **\`ingestion_items\`** — extracted thought candidates per job, with evidence appends tracked via the \`append_thought_evidence\` function

**Security posture:**
- RLS enabled on both tables
- \`service_role ALL\` policies (unconditional)
- \`authenticated SELECT WHERE user_id = auth.uid()\` policies added conditionally if the \`auth\` schema exists (Supabase-only; no-op on vanilla Postgres)
- \`append_thought_evidence\` is \`SECURITY DEFINER\` but restricted to \`service_role\` (no anon/authenticated execute grants)
- \`FOR UPDATE\` row lock in evidence append to prevent lost-update races

**Ops:**
- Partial indexes on \`status='pending'\` / \`status IN ('pending','ready')\` keep pending-queue lookups cheap as the tables grow
- Nullable \`user_id uuid\` columns on both tables; FK to \`auth.users(id) ON DELETE CASCADE\` added conditionally if the auth schema exists
- Fully idempotent (\`IF NOT EXISTS\` / \`CREATE OR REPLACE\` / \`DROP POLICY IF EXISTS\` + \`CREATE POLICY\`)

Replaces the content from the closed PR #98 bundle, with the evidence-append hardening fix (\`f7b4fd7\`) incorporated.

## Why

Stock Open Brain has no primitive for "ingest a document, extract multiple candidate thoughts, let a user review before committing." This schema provides the persistence layer for that workflow. The companion Edge Function (separate PR) handles the extraction logic.

Job-claim semantics live in the Edge Function (\`FOR UPDATE SKIP LOCKED\` pattern) rather than the schema — kept minimal here so operators can use their own claim approach if they wire a custom worker.

## Test plan

- [ ] Apply \`schema.sql\` to a stock Supabase Open Brain project — verify tables, indexes, policies
- [ ] Re-apply — verify no errors (idempotent)
- [ ] Apply to vanilla Postgres (no auth schema) — verify \`NOTICE\` skip of auth.users FK and authenticated policy; RLS still enabled with service_role-only access
- [ ] Call \`append_thought_evidence\` as service_role — succeeds
- [ ] Call as authenticated — fails with \`permission denied\` (confirm grant restriction)
- [ ] Two concurrent \`append_thought_evidence\` calls on same thought_id — verify both appends land without lost-update
- [ ] Verify \`metadata.json\` passes the gate schema


alanshurafa:
Mergeable, no conflicts against `main`. No blockers from my side; ready whenever it reaches the queue.

justfinethanku (approved):
Validated the schema README credit fix and confirmed the branch stays mechanically clean. Ready to merge.

PR #200, opened 2026-04-18
[integrations] Consolidation workers (bio + metadata)

alanshurafa:
## Summary

Two Deno/Supabase Edge Functions for post-import thought quality improvement:

- **\`metadata-norm\`** — re-classifies the \`type\` and \`importance\` fields on accumulated thoughts that were captured pre-enrichment or with weak metadata
- **\`bio\`** — synthesizes a canonical "Who is {subject}" profile thought from biographical captures scoped to that subject

Both workers use three-tier LLM fallback (OpenRouter → OpenAI → Anthropic) with transient-error gating (\`isTransientError\` predicate) so provider hops only fire on 5xx/429/timeout — not on 4xx/auth errors that would waste calls.

Hardened per Wave 2.5 review:
- **\`CONSOLIDATION_MAX_CALLS\` budget** (default 100, 0 = unlimited); preserves \`consolidation_reviewed\` markers on trip
- **AbortController** on every LLM fetch (\`FETCH_TIMEOUT_MS\` default 60s)
- **Prompt-injection defense:** thought content wrapped in \`<thought_content>\` tags (with closing-tag escape via zero-width-space); \`<previous_profile>\` wrapping in bio; system/user role split on all 6 provider calls; \`importance>=6\` output treated as an injection signal (clamps to 0-5)
- **Bio insert via \`.insert()\`** with \`content_fingerprint\` instead of \`upsert_thought\` RPC (fixes the RPC-signature mismatch that dropped \`type\`/\`importance\`/\`source_type\` on first-run profiles)
- **Per-subject profile scoping:** \`findExistingProfile\` filters on \`metadata->>subject\` — profiling multiple subjects no longer overwrites each other's rows
- **UUID types** for \`thought_id\` throughout; dead \`typeof === 'number'\` branches removed
- **b361df1 name-filter** applied across all three source queries in the bio worker

## Why

Open Brain's import paths accumulate thoughts faster than classification can keep up, especially for pre-enrichment backfills (drag-and-drop archives, Takeout dumps). \`metadata-norm\` is a targeted re-classifier for the long tail. \`bio\` is a different shape: synthesize a stable "self" or "about-{person}" anchor document that LLMs can efficiently ground against.

## Test plan

- [ ] POST \`/metadata-norm\` with \`{ "limit": 5 }\` — verify 5 thoughts re-classified, \`consolidation_reviewed: true\` set
- [ ] Set \`CONSOLIDATION_MAX_CALLS=2\`, post with \`limit=5\` — verify 2 processed, loop breaks, response includes \`truncated: { reason: "budget_reached" }\`
- [ ] POST \`/bio\` with \`{ "subject": "self" }\` on first run — verify a profile thought is created with correct type/importance/source_type (not null)
- [ ] POST \`/bio\` twice for \`subject: "Sarah"\` then \`subject: "Alice"\` — verify Sarah's profile is not overwritten by Alice's
- [ ] Kill primary LLM provider (stall), verify fallback fires through to secondary
- [ ] Post with hostile content (prompt-injection attempt) — verify extraction returns clean values, \`importance\` stays 0-5
- [ ] \`deno check\` on both \`index.ts\` files + \`_shared/\`


alanshurafa:
Conflicts with `main`. Before I rebase, should this stay as-is, be reshaped, or be dropped? I can refresh it quickly if you want to keep it.

alanshurafa:
Rebased onto `main` — conflicts cleared. The only conflict was a stray repo-wide markdown-lint commit on this branch that collided with `main`'s own lint cleanup; I dropped that commit. The branch's actual changes are untouched. Now mergeable.

justfinethanku (approved):
Validated both worker entrypoints with deno check and added the missing OB1 community credit header. Ready to merge.

PR #45, opened 2026-03-16
[integrations] REST API gateway for non-MCP clients

alanshurafa:
## Summary
- Adds an optional REST API Edge Function alongside the MCP server
- 8 routes: search, capture, recent, get/update/delete thought, stats, health
- Unlocks ChatGPT Custom Actions, iPhone Shortcuts, Zapier/Make/n8n, web dashboards, and any tool that speaks HTTP
- Same auth as MCP (x-brain-key header or query param)
- Self-contained — no external dependencies beyond @supabase/supabase-js
- Follows the same README structure as the Slack Capture integration (credential tracker, numbered steps, troubleshooting)

## Why
MCP works great for AI assistants (Claude, Cursor), but many tools only speak HTTP. This gateway gives every app, automation, and device a standard REST door into Open Brain without replacing or competing with MCP.

Tested in production with 75K+ thoughts — all 8 routes verified working (semantic search, text search, capture with metadata extraction, CRUD operations, stats).

## Test plan
- [ ] Deploy to Supabase project via `supabase functions deploy open-brain-rest --no-verify-jwt`
- [ ] Test health endpoint: `GET /health`
- [ ] Capture a thought: `POST /capture` with `{"content": "test thought"}`
- [ ] Search semantically: `POST /search` with `{"query": "test", "limit": 5}`
- [ ] List recent: `GET /recent?limit=5`
- [ ] Get by ID: `GET /thought/:id`
- [ ] Update: `PUT /thought/:id` with `{"content": "updated"}`
- [ ] Delete: `DELETE /thought/:id`
- [ ] Verify CORS headers from browser

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #192, opened 2026-04-18
[recipes] Thought enrichment pipeline

alanshurafa:
## Summary

LLM-powered recipe that retroactively classifies existing thoughts with type, importance, quality score, sensitivity tier, and structured metadata (topics, tags, people, action items).

- Supports OpenRouter and Anthropic providers
- Batching, retry, and checkpoint/resume for large backfills
- Sensitivity-tier classification with pattern-based pre-filter
- Separate `backfill-type` and `backfill-sensitivity` scripts for targeted re-runs
- Pure Node.js ESM, no global installs — config via `.env.local`

## Why

Most Open Brain sources hand you raw content. Without enrichment, the `type`, `importance`, `quality_score`, and sensitivity columns from the enhanced-thoughts schema stay null and the thoughts remain hard to retrieve or filter. This recipe is the turn-key way to populate those columns for anyone who's already running Open Brain and has accumulated untyped thoughts.

**Requires** the `schemas/enhanced-thoughts/` columns from [#191](https://github.com/NateBJones-Projects/OB1/pull/191). The README calls that out in Prerequisites.

Part 2 of 12 in the OB1 Alpha Milestone consolidation.

## Test plan

- [ ] Apply `schemas/enhanced-thoughts/schema.sql` first (dependency)
- [ ] Configure `.env.local` with Supabase + one LLM provider key
- [ ] Run `enrich-thoughts.mjs --limit 10` — verify 10 thoughts get classified
- [ ] Re-run with same limit — verify idempotency (already-enriched thoughts skipped)
- [ ] Kill mid-run and resume — verify checkpoint picks up cleanly
- [ ] Run `backfill-sensitivity.mjs --dry-run` — verify pattern matches print but no writes
- [ ] Verify `metadata.json` passes the gate

PR #201, opened 2026-04-18
[integrations] REST API gateway

alanshurafa:
## Summary

Deno/Supabase Edge Function providing a REST API for non-MCP clients — 15 endpoints over thoughts, ingestion, and search.

Hardened per Wave 2.5 review:
- **String IDs** — all thought IDs parsed, validated, and forwarded as strings (not \`Number()\`-coerced). Prevents silent row-identity corruption when IDs cross 2^53 or migrations preserve historical BIGINT values.
- **Robust upsert response parsing** — \`extractThoughtId()\` helper handles scalar, \`{id}\`, \`{thought_id}\`, array-wrapped, and null-returning shapes. No more 500 on a legitimate response.
- **Timing-safe auth** — \`timingSafeEqual()\` XOR-accumulator for \`x-brain-key\` comparison. Early return on missing key; \`.trim()\` handles copy-paste whitespace.
- **CORS allowlist** — \`CORS_ALLOWED_ORIGINS\` env (comma-separated). Legacy default \`*\` preserved for backward compat, but README documents the risk for production.
- **Per-key rate limiting** — 100 req/min default (\`RATE_LIMIT_PER_MIN\` env). SHA-256-hashed keys so logs never leak the raw key. In-memory bucket with \`Retry-After\` response.
- **Ingest proxy hardening** — 60s \`AbortController\` timeout, 1 MB body cap, distinguishes 502 (unreachable), 504 (timeout), invalid-JSON vs forwarded-JSON responses.
- **Sort-column allowlist** — rejects arbitrary column names; prevents schema enumeration and ordering by heavy/PII columns.
- **Error opacity** — top-level catch returns \`{error:"internal_error", code:"GENERIC", error_id:<uuid>}\` and logs full error server-side with the same UUID. No PostgREST constraint names or data values leak to callers.
- **Sensitivity re-detection on update** — \`handleUpdateThought\` runs \`detectSensitivity\` + escalation-only \`resolveSensitivityTier\`. A standard thought rewritten to contain a credit card can no longer stay \`standard\`.

## Why

The stock MCP server is fine for AI clients, but dashboards, scripts, and webhooks need HTTP. This gateway mirrors the core MCP's capture/search/list/update surface over REST with the same auth model (\`x-brain-key\`) — so any client that can do HTTPS + headers can talk to an Open Brain.

Replaces the content from the closed PR #45 bundle with the Wave 2.5 security hardening pass incorporated.

## Test plan

- [ ] POST \`/capture\` with valid key — thought created, ID returned as string
- [ ] POST \`/capture\` with wrong key — 401 returned in constant time (no timing-based key-length leak)
- [ ] Flood 150 requests in 60s with same key — verify 101st gets 429 + \`Retry-After\`
- [ ] GET \`/thoughts?sort=embedding\` — verify 400 (sort column not in allowlist)
- [ ] GET \`/thoughts?sort=created_at\` — verify 200 + sorted results
- [ ] POST \`/update\` changing content to include \`4111-1111-1111-1111\` on a \`standard\` thought — verify tier escalates to \`personal\` or \`restricted\`
- [ ] Unset \`CORS_ALLOWED_ORIGINS\` → verify \`*\` still honored with README warning
- [ ] Set \`CORS_ALLOWED_ORIGINS="https://app.example.com"\` → verify other origins rejected
- [ ] \`deno check index.ts\` + \`_shared/*.ts\`


alanshurafa:
Conflicts with `main`. This is tied to the design discussion in #192; I'll rebase once there's a direction there.

alanshurafa:
Rebased onto `main` — conflicts cleared. The only conflict was a stray repo-wide markdown-lint commit on this branch that collided with `main`'s own lint cleanup; I dropped that commit. The branch's actual changes are untouched. Now mergeable.

alanshurafa:
Both #201 and #202 are single-tenant, shared-key gateways over a service-role backend (RLS bypassed by design). Both now use constant-time key comparison, env-only secrets, and header-only auth — I dropped the `?key=` query param from #201 (it leaked the key into access logs) to match #202. Both READMEs now document the trust model explicitly: single-tenant by design, high-entropy key required (rate limiting is best-effort), and the service-role/RLS-bypass boundary. The auth surface is sound for the intended self-hosted use — ready for your review whenever.

justfinethanku (approved):
Audited for maintainer merge lane: branch is mergeable, Alan already documented the single-tenant trust model clearly, local  passes, and I added the standard Community Contribution badge plus creator credit to the README so the contribution convention is preserved.

justfinethanku (approved):
Audited for the maintainer merge lane: the branch is mergeable, Alan already documented the single-tenant trust model clearly, local Deno type-checking for integrations/rest-api passes, and I added the standard Community Contribution badge plus creator credit to the README so the contribution convention is preserved.

PR #229, opened 2026-04-23
[integrations] delete_thought MCP

txcfi-scott:
## What it does

Adds a standalone MCP Edge Function that exposes a single new tool: `delete_thought(id)`.

The core Open Brain MCP server does not expose a delete path. As a result, thoughts accumulate forever — there is no way for an AI client to remove a test entry, a duplicate, or something captured in error without dropping into the Supabase SQL editor.

This integration deploys a second Edge Function with exactly one tool, and the core `server/index.ts` is not modified.

## Safety

- **Hard delete.** The row is removed from `thoughts`. Recovery depends on your Supabase project's backup tier (daily backups on paid tiers; Point-in-Time Recovery on higher tiers).
- **Pre-flight fetch.** If the id does not exist, the server returns a clear `Thought not found: <id>` instead of silently succeeding.
- **Audit hook.** If the companion `schemas/thought-audit` schema is installed, the README documents how to extend this function to write an audit row (with prior content + metadata) before the delete so the operation is recoverable from the audit trail. Left out of the base integration to keep its dependencies minimal.

## What it requires

- Working Open Brain setup (just the `thoughts` table from the getting-started guide).
- Supabase CLI for deployment.

No LLM provider is required — this tool does not read or write embeddings.

## Tested

Deployed against a live Supabase Open Brain instance. Verified:

- Deleting an existing UUID returns `Deleted thought <id> (prior content length: N chars).` and the row disappears from the table.
- Deleting a non-existent UUID returns `Thought not found: <id>` with `isError: true`.
- The existing `search_thoughts`, `list_thoughts`, and `capture_thought` tools from the core server continue to work — this integration is additive and side-effect-free for existing connectors.

## Files touched

Only `integrations/delete-thought-mcp/` — stays within the contribution scope check.

## Attribution

Adapted from a multi-participant capture design used across live Claude / ChatGPT / Codex sessions.


alanshurafa:
Thanks for the contribution. The core gap is real — the MCP server has no delete path, and `delete_thought` fills it cleanly: a pre-flight existence check (so the caller gets a clear not-found), a confirmation response, and a standalone connector. The README is honest that this is a hard delete.

One design question worth a maintainer decision: this exposes a hard delete to an AI client, with recovery only via Supabase backups and audit left optional (the README points to `schemas/thought-audit` / #230 as an opt-in extension). An LLM calling `delete_thought` on a wrong or hallucinated UUID is unrecoverable. Worth deciding whether OB1's delete path should be hard delete (this PR) or default to a recoverable soft delete with built-in audit. The code itself is correct for the hard-delete design it implements.

Needs product decision on the delete-path design; recommend maintainer review.

txcfi-scott:
@alanshurafa Alan's question is load-bearing — you're right that exposing a hard delete to an AI client with backup-only recovery is a design decision, not just a process concern. Having thought about it more, I think the better default for OB1's delete path is a recoverable soft-delete with built-in age-out.

**Preferred design:** add a `deleted_at TIMESTAMPTZ` column to `thoughts`. The `delete_thought` tool sets `deleted_at = NOW()` rather than issuing a `DELETE FROM`. All read queries — `search_thoughts`, `list_thoughts`, `capture_thought` responses, stats — filter `WHERE deleted_at IS NULL`, so normal callers see no change in semantics. A scheduled age-out job (Supabase cron or an external timer) hard-deletes rows where `deleted_at < NOW() - INTERVAL '30 days'` — open to a different retention window if there's prior art in the repo. A new `recover_thought(id)` RPC sets `deleted_at = NULL` for any row still within the window.

The practical gain: an LLM calling `delete_thought` on a wrong or hallucinated UUID is recoverable for 30 days without needing to touch Supabase backups or an optional audit schema. The cost is real — this is bigger than a single connector change. It requires a new migration, touches every read query in the core server, and needs the age-out job placed somewhere. Not a trivial scope expansion.

**Two paths for sequencing:** (a) revise this PR in-place to implement soft-delete + age-out + recovery — bigger commit but a cleaner story for the delete path; or (b) merge or leave this PR as-is and open a follow-up PR that adds soft-delete as a first-class schema feature, superseding or wrapping this connector. Nate, which sequencing makes more sense from a release cadence standpoint?

Thanks again, Alan — this was the kind of question that actually changed the design.

JodieM:
Any update on this please? I would love to be able to delete thoughts no longer relevant, but I do not need to go the full dashboard app route

txcfi-scott:
Hi Jodie — thanks for jumping in. Genuinely useful to hear from someone who wants the capability but doesn't want to stand up a whole dashboard for it.

Quick status: the function in this PR works. What's pending is the maintainer-side call on sequencing — revise in place, or merge and follow up with the soft-delete refinement. I sketched the soft-delete design in the 2026-05-21 comment further up, so I won't re-litigate. How I got there is worth a few lines, though, because it touches your use case.

Earlier in 2026 I imported about 1,939 thoughts from Fastmail via IMAP. The April audit found roughly 17% were straight noise — marketing, transactional, automated. Unglamorous lesson: bytes you might want tomorrow shouldn't disappear today. Then Alan flagged the LLM angle on 2026-05-20 — an AI calling delete on a hallucinated UUID is unrecoverable, a worse failure mode than a missed cleanup. Soft-delete with a 30-day window covers both, and a human checkpoint already exists in the rhythm: Pinky (the curation agent — and yes, she curates the Brain; NARF, I didn't pick the names so much as fail to resist them) re-embeds null vectors and spot-checks weekly, then runs a monthly full re-slice — ledger regenerated, distribution deltas surfaced, links built at 0.85–0.95 cosine, merges at 0.95 and above. The 30-day age-out is one full maintenance cycle, not an arbitrary number. The April sprint itself — 3,146 thoughts in 30 batches, 69% pass / 22% reclassify / 17% dup-candidate / 1% low-value — produced zero auto-deletions. Everything destructive routes back through me.

The part I keep wondering about, and I want to flag as speculation rather than a roadmap claim: what we're sketching looks a little like how brains use sleep. Monthly consolidation is already the engineering version — dedup, abstraction, link-building during idle time. The more speculative version is whether the brain should also dream — synthesize patterns across captured traces, surface emergent abstractions, and let truly low-signal traces fade as part of the rhythm rather than as a separate delete event. I don't know if "dreaming" is the right metaphor or whether it leads anywhere useful, but it feels like a question worth opening: should a memory system that's already consolidating do something more reverie-like during idle time, and would that reframe how we think about deletion at all? Curious where the room lands.

Short version for you, Jodie: the capability is real and works today. The hard-vs-soft call is what's holding the merge, and that one isn't mine to make. Nothing you need to do — but if any of this lands, I'd love your read.

txcfi-scott:
@justfinethanku — updating my own earlier sequencing question so you have a clear decision to make rather than an open fork.

I'm going to revise this PR in place to implement soft-delete. The hard-delete connector that's currently here will be replaced with:

- `deleted_at TIMESTAMPTZ` column on `thoughts`
- `delete_thought` sets `deleted_at = NOW()` rather than issuing `DELETE FROM`
- All read paths (`search_thoughts`, `list_thoughts`, `stats`) filter `WHERE deleted_at IS NULL`
- `recover_thought(id)` RPC to un-delete within the window
- Supabase cron job for hard-deletion of rows older than 30 days

Alan's 2026-05-20 review identified the LLM-calling-delete-on-a-hallucinated-UUID failure mode as the design-level concern. Soft-delete addresses it structurally. The 30-day window also aligns with how curation actually runs in practice — one full maintenance cycle before anything is permanently gone.

@JodieM — this should cover your use case directly: `delete_thought` stays available, and soft-delete means an accidental or hallucinated delete is recoverable rather than permanent.

I'll revise the PR branch over the coming weeks. If there's a reason to keep the hard-delete path as-is and add soft-delete as a follow-up instead, let me know and I'll go that route — but absent direction, I'll revise in place.

justfinethanku (approved):
Maintainer pass complete. Verified the integration README now carries the OB1 community credit, refreshed the branch against current main, and ran `deno check --config integrations/delete-thought-mcp/deno.json integrations/delete-thought-mcp/index.ts` on the updated head.

PR #228, opened 2026-04-23
[integrations] update_thought MCP with if_unchanged_since

txcfi-scott:
## What it does

Adds a standalone MCP Edge Function that exposes a single new tool: `update_thought(id, content?, metadata_patch?, if_unchanged_since?)`.

The core `server/index.ts` is curated and does not expose an update path. This integration adds one without touching the core server — it deploys as its own Supabase Edge Function and is registered as a separate custom connector alongside the main Open Brain connector.

## Why it matters

Once more than one agent writes to the same Open Brain (Claude Desktop + Codex + a background worker, or a shared brain across teammates), last-write-wins silently drops concurrent edits. Optimistic concurrency is the cheapest fix:

1. Caller reads a thought, notes its `updated_at` (T0).
2. Caller issues `update_thought` with `if_unchanged_since = T0`.
3. If the stored `updated_at` is still `<= T0`, the update proceeds. If another writer has advanced it past T0, the server rejects with `STALE_READ` and the caller can re-fetch and retry.

Omitting `if_unchanged_since` keeps the old last-write-wins behaviour — fully backward compatible.

## What it requires

- Working Open Brain setup (the `thoughts` table + the `update_updated_at` trigger from the getting-started guide).
- Supabase CLI for deployment.
- OpenRouter API key — only exercised when callers pass `content` (re-embedding path).

## Tested

Deployed against a live Supabase Open Brain instance. Verified:

- `metadata_patch`-only updates succeed without an OpenRouter key.
- `content` updates regenerate the embedding and the row's `updated_at` advances.
- Two concurrent writers, the second passing `if_unchanged_since` from its pre-mutation read, receive a clear `STALE_READ` error once the first writer has committed.

## Files touched

Only `integrations/update-thought-mcp/` — stays within the contribution scope check.

## Attribution

Adapted from a multi-participant capture design used across live Claude / ChatGPT / Codex sessions. Released here as a standalone integration so any Open Brain user can opt in without modifying the core server.

PR #230, opened 2026-04-23
[schemas] thought_audit table for multi-participant attribution

txcfi-scott:
## What it does

Adds a `thought_audit` table plus an `author_session_id` convention, so **who changed what, when** becomes answerable once more than one participant writes to the same Open Brain.

**The problem.** Once you have Claude Desktop, Codex, a ChatGPT API worker, and a nightly importer all writing to the same `thoughts` table, updates and deletes are effectively anonymous — nothing records which session or which agent did what. For a personal brain that is fine. For a shared brain, or for debugging "why did this thought disappear?", it is painful.

**The solution.** An append-only audit table that records every mutation, plus a small convention (`metadata.author_session_id`) that lets you cluster all writes from a single agent session.

## How it stays safe

- **Additive only.** No existing thoughts columns are altered or dropped. No existing functions are replaced. Fully idempotent — safe to re-run.
- **No foreign key.** `thought_audit.thought_id` deliberately has no FK to `thoughts(id)` so audit rows survive deletion of their subject (the most important audit event to preserve).
- **INSERT-only grants.** `service_role` gets `SELECT, INSERT` on the audit table — never `UPDATE` or `DELETE`. Nothing downstream can rewrite history without an explicit, reviewed migration that temporarily grants DELETE.
- **Best-effort writes.** The README documents the audit-insert pattern as fire-and-forget: a failure to record an audit row must never block the main capture / update / delete.

## Example audit trail

```
t=10:00  capture  source=claude-desktop     session=sess-AAA  thought-123
t=10:05  update   source=claude-desktop     session=sess-AAA  thought-123 (metadata patch)
t=11:30  update   source=chatgpt-api        session=sess-BBB  thought-123 (content rewrite)
t=15:00  delete   source=codex-cli          session=sess-CCC  thought-123 (previous_content preserved)
```

One `SELECT * FROM thought_audit WHERE thought_id = '...' ORDER BY created_at` returns the full provenance.

## What it requires

- Working Open Brain setup (just the `thoughts` table).

No additional services. No LLM calls.

## Companion integrations

This schema is storage-only — the actual audit writes live in the mutation tools. Two companion integrations in separate PRs wire them up:

- `integrations/update-thought-mcp` (#228) — the update tool; its README documents the audit-insert hook.
- `integrations/delete-thought-mcp` (#229) — the delete tool; its README documents the audit-insert hook with prior-content preservation.

If those PRs are still open when this one is reviewed, the README examples here are the authoritative reference. The schema itself has zero runtime dependency on the integrations — you can install it and write audit rows from your own mutation code if you prefer.

## Tested

- Ran `schema.sql` and `author-session-id.sql` twice against a fresh and a populated Supabase Open Brain. Both runs complete idempotently with no errors.
- Attempted `UPDATE thought_audit ...` and `DELETE FROM thought_audit WHERE ...` as `service_role`; both correctly reject with "permission denied for table thought_audit" — confirming the append-only grant works.
- Wrote three audit rows (capture / update / delete) and queried by `thought_id`, `author_session_id`, and `created_at`; all three indexes get used.
- Dropped a thought, confirmed the `delete` audit row remains with `diff.previous_content` preserved.

## Files touched

Only `schemas/thought-audit/` — stays within the contribution scope check.


alanshurafa:
Thanks for the contribution. This is a well-designed audit schema — the deliberate choices are sound: append-only (`service_role` gets `SELECT` / `INSERT` only, never `UPDATE` / `DELETE`), the non-FK on `thought_id` so audit rows outlive deleted thoughts, idempotent statements, and strictly additive with no change to the `thoughts` table structure.

Two notes for the maintainer. The audit table is storage-only with no trigger, so coverage depends on every mutation tool writing a row — it pairs naturally with this contributor's #228 (`update_thought`) and #229 (`delete_thought`). And `author_session_id` is a convention on `thoughts.metadata`, so it is worth documenting consistently if adopted. Recommend maintainer review.

txcfi-scott:
@alanshurafa Thanks. The non-FK on `thought_id` was the most deliberate choice in the schema — a foreign key would cascade-delete exactly the audit rows that matter most (the ones documenting a deletion). You're right that `author_session_id` needs a conventions doc; its shape is currently determined by the upstream callers in #228/#229 rather than enforced at the schema layer. If this lands with those two I'll follow up with a short doc covering the expected format and semantics.

alanshurafa:
The cascade-delete point is exactly why the non-FK is correct — audit rows documenting a deletion are the ones you least want a foreign key to remove. A short `author_session_id` conventions doc landing with #228/#229 would close the loop. Ready for maintainer from my side.

— Alan

justfinethanku (approved):
Maintainer review: Alan's triage says the append-only schema design is sound and ready for maintainer. I updated the branch, added the repo community badge/creator credit convention, ran markdownlint locally, and GitHub label/lint checks are passing.

PR #232, opened 2026-04-23
[extensions] Add RLS policies to family-calendar

AnkitClassicVision:
## Summary
- Enables Row Level Security on all three family-calendar tables (`family_members`, `activities`, `important_dates`)
- Policies: `auth.uid() = user_id` for user access + `service_role` bypass
- Updates the README note to reflect the shift: family-calendar data is personal enough that tenant isolation is worth turning on from the start

## Why
Extensions 1-3 were originally framed as single-user systems where RLS isn't strictly required until Extension 4 (Meal Planning). But family-calendar data (kids' schedules, medical appointments, birthdays) is sensitive enough that it's worth defending at the database layer from day one — especially as users may eventually connect multiple clients or share limited views.

The RLS shape mirrors Extension 4's pattern exactly, so the learning path still flows naturally: users see RLS once here, then learn *why* it matters for shared household access in Extension 4.

## Test plan
- [ ] Apply schema.sql to a fresh Supabase project
- [ ] Verify RLS is enabled on all three tables (`SELECT relrowsecurity FROM pg_class WHERE relname IN ('family_members','activities','important_dates')`)
- [ ] Verify a non-service-role JWT with `auth.uid()` = user_X can only read/write their own rows
- [ ] Verify service_role bypasses correctly


alanshurafa:
This branch conflicts with current `main`. Recommend author refresh — a rebase should clear it. One note for the maintainer: the RLS policy shape will want a review against the rest of the extension path before merge, since it changes data-access rules.

PR #227, opened 2026-04-23
[schemas] Fix invalid || concatenation in typed-reasoning-edges COMMENT

spiritualsystems:
## Contribution Type

- [x] Schema (`/schemas`)
- [ ] Recipe (`/recipes`)
- [ ] Dashboard (`/dashboards`)
- [ ] Integration (`/integrations`)
- [ ] Skill (`/skills`)
- [ ] Repo improvement (docs, CI, templates)

## What does this do?

Fixes a SQL syntax error in `schemas/typed-reasoning-edges/schema.sql` (introduced in #208) that prevents the migration from applying. The `COMMENT ON FUNCTION public.thought_edges_upsert IS ...` statement uses `||` string concatenation across three adjacent string literals, but PostgreSQL requires the right-hand side of `COMMENT ON ... IS` to be a single string literal — not an expression. Flattens the three pieces into one literal. No functional change to the function itself.

## Bug

At [schemas/typed-reasoning-edges/schema.sql#L257-L260](https://github.com/NateBJones-Projects/OB1/blob/main/schemas/typed-reasoning-edges/schema.sql#L257-L260):

```sql
COMMENT ON FUNCTION public.thought_edges_upsert IS
  'Insert or (on duplicate key) bump support_count + refresh temporal bounds. ' ||
  'Call via POST /rpc/thought_edges_upsert. Use instead of a plain INSERT when ' ||
  'you want repeated classifications of the same pair to accumulate evidence.';
```

Per the [Postgres grammar for `COMMENT`](https://www.postgresql.org/docs/current/sql-comment.html), `IS` takes a `text` literal, not a general expression — concatenation operators aren't parsed here. None of the other `COMMENT ON` sites in this file have the issue; this is the only affected statement.

### Reproduce

From a clean Supabase project with the rest of Open Brain applied:

```bash
# Stage this schema as a migration and push
supabase db push
```

Result on `main`:

```
ERROR: syntax error at or near "||" (SQLSTATE 42601)
At statement: 31
  COMMENT ON FUNCTION public.thought_edges_upsert IS
    'Insert or (on duplicate key) bump support_count + refresh temporal bounds. ' ||
    ...
```

The transaction rolls back and none of the typed-reasoning-edges objects (`thought_edges` table, `thought_edges_upsert` RPC, entity-edge temporal columns) get created.

## Fix

Collapse the three concatenated literals into one. The surrounding code is untouched; other `COMMENT ON` lines in the file are not reformatted.

```sql
[...]

PR #233, opened 2026-04-23
[schemas] Fix typed-reasoning-edges COMMENT syntax error

AnkitClassicVision:
## Summary
The `COMMENT ON FUNCTION public.thought_edges_upsert IS` clause uses runtime string concatenation (`'...' || '...' || '...'`), which Postgres rejects. `COMMENT` requires a single string literal, not an expression.

## Repro
Apply the schema against Supabase:
```bash
supabase db query --linked --file schemas/typed-reasoning-edges/schema.sql
```
Fails with:
```
ERROR:  42601: syntax error at or near "||"
LINE 258:   'Insert or (on duplicate key) bump support_count + refresh temporal bounds. ' ||
```

## Fix
Replaced `||` with SQL-standard adjacent-string-literal concatenation — the parser joins adjacent `'a' 'b'` into `'ab'` when separated only by whitespace, so the multi-line source stays readable while producing a valid single literal.

```sql
-- before
COMMENT ON FUNCTION public.thought_edges_upsert IS
  'Insert or (on duplicate key) ... ' ||
  'Call via POST ... ' ||
  'you want ...';

-- after
COMMENT ON FUNCTION public.thought_edges_upsert IS
  'Insert or (on duplicate key) ... '
  'Call via POST ... '
  'you want ...';
```

## Test plan
- [x] Hit the original error by applying schema as-is
- [x] Apply patched schema against a Supabase Pro project — applies cleanly
- [x] Verify `\df+ public.thought_edges_upsert` shows the merged description


alanshurafa:
The intended change here — the `typed-reasoning-edges` COMMENT syntax fix in `schemas/typed-reasoning-edges/schema.sql` — has already landed cleanly as a one-file diff in #227 and #303.

This PR also includes roughly fifteen unrelated files (a working `AGENTS.md`, `.vscode/` settings, `.claude/` settings, and personal `docs/brainstorming/` notes) that look like they were committed by accident from another project's working tree. Recommend close as superseded — #227 and #303 cover the actual fix. If anything here differs from those, a fresh single-file PR scoped to just the schema change would be the cleaner path.

PR #243, opened 2026-04-26
[server] return JSON-RPC error envelopes on auth failure

txcfi-scott:
## Problem

The reference MCP server returns a bare HTTP 401 on auth failure:

```ts
return c.json({ error: "Invalid or missing access key" }, 401, corsHeaders);
```

Strict MCP clients that follow the JSON-RPC 2.0 wire spec — Codex CLI, Claude Code, and anything else that distinguishes transport faults from application errors — interpret a bare HTTP 4xx as a transport-level failure and tear the connection down rather than surfacing the auth error to the application layer. The practical consequence is that scenarios like "wrong key — prompt the user to update it" or "stale cached key — refetch and retry" can't be handled gracefully: the client sees a dead connection, not a recoverable error.

This bug bites anyone wiring a fresh OB1 deployment up to a strict client and rotating keys (or fat-fingering the first one).

## Solution

Return HTTP 200 with a JSON-RPC 2.0 error envelope. The transport stays healthy, and the JSON-RPC layer carries the failure as an application-level error the client can handle:

```json
{ "jsonrpc": "2.0", "error": { "code": -32001, "message": "Unauthorized: missing or invalid authentication." }, "id": <inbound id or null> }
```

`-32001` lives in JSON-RPC's `-32099..-32000` "implementation-defined server error" range and is the conventional code MCP servers use for "Unauthorized" in the wild. The inbound request `id` is best-effort extracted and echoed back so the response correlates with the call; malformed or missing bodies fall back to `id: null`.

## Backward compatibility

- **Success path unchanged.** Authenticated requests behave exactly as before — same StreamableHTTPTransport handoff, same Accept-header patch, same CORS.
- **Lenient clients still work.** A client that doesn't validate JSON-RPC envelopes still gets a 200 with a parseable JSON body containing `error`. It can read the message or ignore it.
- **Strict clients now recover gracefully** instead of dying on a transport fault.

## Edge cases handled

- **Bodyless methods (GET / HEAD / DELETE):** body read is skipped, `id` defaults to `null`.
- **Malformed JSON body:** parse failure swallowed, `id` defaults to `null`.
- **Non-JSON-RPC body shape:** missing/non-id-typed `id` falls back to `null` (only `string | number | null` are accepted, per the JSON-RPC 2.0 spec).
- **CORS preflight (`OPTIONS`):** unchanged — still returns `ok` 200 with CORS headers before auth runs.
- **Body consumed on auth failure:** safe — on the rejection path we return immediately and never forward to the transport, so consuming the body to read the id has no downstream effect.

## Implementation

Three small helpers added next to the existing `corsHeaders` block:

- `readBodyText(req)` — best-effort body read, returns `null` on bodyless methods or read failure.
- `extractJsonRpcId(bodyText)` — best-effort JSON-RPC id extraction, returns `null` on anything malformed.
- `unauthorizedResponse(id)` — builds the HTTP 200 + JSON-RPC error envelope with CORS headers attached.

The auth check in `app.all("*", ...)` calls those helpers in place of the bare `c.json(...)` 401.

## Tested

- `deno check server/index.ts` — no type errors.
- Reviewed against the original report from the downstream MCP host (Codex CLI) where bare-401 originally tripped the transport-fault path. The same envelope-shaped response is what fixed the symptom downstream.
- Verified by inspection that the success path, the Accept-header patch, CORS preflight, and the StreamableHTTPTransport handoff are byte-for-byte unchanged.

A wire-level diagnosis of why bare 4xx breaks strict MCP hosts (and why a JSON-RPC envelope is the right fix) was written up separately during the downstream investigation. Happy to share or adapt the relevant parts as a `docs/` note if useful — let me know.

## Files touched

- `server/index.ts` — only file changed. +85 / -1.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -370,6 +370,83 @@ const corsHeaders = {
   "Access-Control-Allow-Methods": "GET, POST, OPTIONS, DELETE",
 };
 
+// JSON-RPC error code for unauthorized requests.
+// Per the JSON-RPC 2.0 spec, the range -32099 to -32000 is reserved for
+// implementation-defined server errors. -32001 is the conventional
+// "Unauthorized" code used by MCP clients/servers in the wild.
+//
+// Why a JSON-RPC envelope (HTTP 200) instead of a bare HTTP 401?
+// Strict MCP hosts (Codex CLI, Claude Code) treat bare HTTP 4xx responses
+// as transport-level failures and tear the connection down rather than
+// surfacing the failure to the application layer. Wrapping the auth
+// rejection in a JSON-RPC error keeps the connection alive and lets
+// clients recover (e.g. prompt the user for a new key, refetch a stale
+// cache) instead of dying.
+const JSON_RPC_UNAUTHORIZED_CODE = -32001;
+const UNAUTHORIZED_MESSAGE = "Unauthorized: missing or invalid authentication.";
+
+/**
+ * Read the request body as text without consuming the original request's
+ * body stream for downstream handlers. Returns null on bodyless methods
+ * or read failure.
+ */
+async function readBodyText(req: Request): Promise<string | null> {
+  if (req.method === "GET" || req.method === "HEAD" || req.method === "DELETE") {
+    return null;
+  }
+  try {
+    return await req.text();
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Best-effort extraction of the JSON-RPC `id` from a raw request body.
+ * Returns null when the body is missing, not JSON, or not a JSON-RPC
+ * shape with an id. Per the JSON-RPC 2.0 spec, id may be a string,
+ * number, or null — we preserve any of those; anything else becomes null.
+ */
+function extractJsonRpcId(bodyText: string | null): string | number | null {
+  if (!bodyText) return null;
+  try {
+    const parsed = JSON.parse(bodyText);
+    if (parsed && typeof parsed === "object" && "id" in parsed) {
+      const id = (parsed as { id: unknown }).id;
+      if (typeof id === "string" || typeof id === "number" || id === null) {
+        return id;
+      }
+    }
+  } catch {
+    // fall through — malformed body
+  }
+  return null;
+}
+
+/**
+ * Build a JSON-RPC 2.0 error envelope response for auth failures.
+ * Returns HTTP 200 — the JSON-RPC layer expresses the error so that
+ * strict MCP clients keep the connection alive instead of treating
+ * the failure as a transport-level fault.
+ */
+function unauthorizedResponse(id: string | number | null): Response {
+  const body = {
+    jsonrpc: "2.0",
+    error: {
+      code: JSON_RPC_UNAUTHORIZED_CODE,
+      message: UNAUTHORIZED_MESSAGE,
+    },
+    id,
+  };
+  return new Response(JSON.stringify(body), {
+    status: 200,
+    headers: {
+      "Content-Type": "application/json",
+      ...corsHeaders,
+    },
+  });
+}
+
 const app = new Hono();
 
 // CORS preflight — required for browser/Electron-based clients (Claude Desktop, claude.ai)
@@ -381,7 +458,14 @@ app.all("*", async (c) => {
   // Accept access key via header OR URL query parameter
   const provided = c.req.header("x-brain-key") || new URL(c.req.url).searchParams.get("key");
   if (!provided || provided !== MCP_ACCESS_KEY) {
-    return c.json({ error: "Invalid or missing access key" }, 401, corsHeaders);
+    // Return a JSON-RPC 2.0 error envelope (HTTP 200) instead of a bare
+    // HTTP 401 so strict MCP hosts treat this as an application-level
+    // error rather than a transport fault and keep the connection alive.
+    // Best-effort echo of the inbound request id keeps the response
+    // correlated; malformed/missing bodies fall back to id: null.
+    const bodyText = await readBodyText(c.req.raw);
+    const id = extractJsonRpcId(bodyText);
+    return unauthorizedResponse(id);
   }
 
   // Fix: Claude Desktop connectors don't send the Accept header that


alanshurafa:
Thanks for the contribution. Reviewed this closely since it touches the auth path.

The security side checks out: the auth predicate (`!provided || provided !== MCP_ACCESS_KEY`) is unchanged, so no bypass is introduced. The error envelope leaks nothing — `code` and `message` are constants, and the echoed `id` is the client's own request id, correctly restricted to `string | number | null` per the JSON-RPC 2.0 spec. `readBodyText` is only reached on the reject path, so consuming the body there has no downstream effect.

The open question is the HTTP 200 status for an auth failure. It deviates from the usual convention (auth failures as 401), and anything that branches on status code — monitoring, a WAF, rate-limit-on-401 rules — would stop registering these as failures. The MCP-transport rationale (strict hosts treating a bare 4xx as a connection-level fault) is plausible and worth substantiating: do you have a concrete repro from Codex CLI or Claude Code showing the connection teardown on a 4xx? That would give the maintainer a clear basis to weigh the tradeoff. Happy to help verify against a real MCP client if useful.

Recommend author refresh with that evidence; the status-code convention itself is a maintainer call.

txcfi-scott:
@alanshurafa You asked for a concrete repro of Codex CLI teardown on a bare 4xx. Here it is, with method.

### Captured artifact — pre-fix wire behavior (Approach B: mock server)

Since rolling back the live deployment would create noise for real users, I reproduced the pre-fix wire shape with a minimal mock server (Python, saved to `/Volumes/Danger/claude-scratch/tmp/mock-prefix-401-v2.py` for reproducibility) that returns exactly the pre-fix response for any POST:

```
HTTP/1.0 401 Unauthorized
Content-Type: application/json
Content-Length: 42
Access-Control-Allow-Origin: *

{"error": "Invalid or missing access key"}
```

That body matches the BEFORE side of the PR diff exactly: `c.json({ error: "Invalid or missing access key" }, 401, corsHeaders)`. (The diff's own hunk header is `@@ -381,7 +458,14 @@`; the `c.json(…, 401, …)` line is in the BEFORE side.)

**Codex CLI (v0.125.0) against the bare-401 endpoint:**

```
2026-05-20T14:24:33Z ERROR rmcp::transport::worker: worker quit with fatal:
  Transport channel closed, when Deserialize(Error(
    "data did not match any variant of untagged enum JsonRpcMessage",
    line: 0, column: 0))

2026-05-20T14:24:51Z ERROR codex_core::tools::router:
  error=resources/list failed: failed to get client:
  MCP startup failed: handshaking with MCP server failed:
  Send message error Transport [rmcp::transport::worker::WorkerTransport<
    rmcp::transport::streamable_http_client::StreamableHttpClientWorker<
      codex_rmcp_client::http_client_adapter::StreamableHttpClientAdapter>>]
  error: Deserialize error: data did not match any variant of
  untagged enum JsonRpcMessage, when send initialize request

mcp: mock-prefix-401/list_mcp_resources started
mcp: mock-prefix-401/list_mcp_resources (failed)
```

Note that the proximate trigger is the deserializer failing on the non-JSON-RPC body, not the 401 status itself — see §7 for the untested 401+envelope variant.

The bare HTTP 401 body `{"error":"Invalid or missing access key"}` is not parseable as a JSON-RPC envelope (`JsonRpcMessage`), so `rmcp`'s transport worker closes the channel immediately during the `initialize` handshake. The failure surfaces to the user as MCP startup failed, with no indication of _why_ (wrong key, no key, service down — all look the same).

**Full log:** `/Volumes/Danger/claude-scratch/tmp/codex-teardown-prefix.log`

---

**Codex CLI (v0.125.0) against the post-fix endpoint (live deployment, wrong key):**

```
HTTP/2 200
content-type: application/json
...

{"jsonrpc":"2.0","error":{"code":-32001,"message":"Unauthorized: missing or
 invalid authentication. Provide both Authorization Bearer token and
 x-brain-key (header or query param)."},"id":1}

2026-05-20T14:28:16Z ERROR codex_core::tools::router:
  error=resources/list failed: failed to get client:
  MCP startup failed: handshaking with MCP server failed:
  JSON-RPC error: -32001: Unauthorized: missing or invalid authentication.
  Provide both Authorization Bearer token and x-brain-key
  (header or query param).

mcp: postfix-wrongkey/list_mcp_resources started
mcp: postfix-wrongkey/list_mcp_resources (failed)
```

The post-fix path reaches the same `(failed)` outcome for an invalid key — the session still doesn't open — but the error is now a structured `JSON-RPC error: -32001: Unauthorized` message rather than a deserialize panic. The host has a typed, correlated, human-readable error it can surface; the pre-fix path produced a transport-level crash with no semantic content.

**Full log:** `/Volumes/Danger/claude-scratch/tmp/codex-teardown-postfix.log`

---

### 1. The convention concern is right

Returning HTTP 200 on an authentication failure is a deviation from the usual REST/HTTP contract. Anything that branches on status code — uptime probes, WAF rules, rate-limit-on-401 policies, log dashboards counting `4xx` as auth errors — won't see these as failures anymore. That's a real cost and worth naming. The auth predicate itself is unchanged, so the change is purely about how the rejection is shaped on the wire; it is not a relaxation of who gets in.

### 2. Why the deviation was made

MCP's Streamable HTTP transport (per the MCP spec) is JSON-RPC 2.0 layered over HTTP. For Codex CLI, when a non-2xx response arrives during the `initialize` handshake, `rmcp`'s transport worker closes the channel — the deserialization error logged above is consistent with this mechanism (inferred from log ordering; rmcp source not read). A bare 401 therefore matches the symptom I observed downstream: Codex CLI reports an MCP connection or initialization error rather than an auth error, with no actionable diagnostic for the user.

For Claude Code, I retested with the same mock server (v2: POST→401, GET→404) and observed a different failure pathway. Claude Code v2.1.144 uses an OAuth-first handshake: it POSTs to `/mcp` (receives the bare 401), then probes multiple OAuth metadata endpoints (all returned 404), then attempts dynamic client registration (`POST /register`, also 401). After exhausting all OAuth paths, the server is silently marked as failed/unavailable — the session continues without it, and no error is shown to the user unless they explicitly ask which servers failed to connect. The mock server request log for this test:

```
[mock-v2] POST /mcp -> 401
[mock-v2] GET /.well-known/oauth-authorization-server -> 404
[mock-v2] GET /.well-known/openid-configuration -> 404
[mock-v2] GET /.well-known/oauth-protected-resource/mcp -> 404
[mock-v2] GET /.well-known/oauth-protected-resource -> 404
[mock-v2] POST /register -> 401
```

Claude Code then reported: `mock-prefix-401-v2 failed to connect (HTTP at http://localhost:8766/mcp)`. No tools from this server appeared in the session.

Both clients therefore fail on a bare 401, but via different mechanisms. The 200 + JSON-RPC error envelope keeps the transport alive long enough for the host to deliver code `-32001` at the application layer. The artifact above shows this is empirically true for Codex CLI; for Claude Code the mechanism is different — the 200 response would satisfy the initial POST check and bypass the OAuth-exhaustion failure path, surfacing the error as a typed JSON-RPC message rather than a silent drop.

### 3. Real wire trace — current (post-fix) behavior

`curl -i` against the live deployment with an invalid key (also shown inline above, in the post-fix Codex run):

```
$ curl -i -X POST 'https://ybsfpdnywkiookmmruke.supabase.co/functions/v1/open-brain-mcp/mcp' \
    -H 'Content-Type: application/json' \
    -H 'x-brain-key: invalid-test-key-not-real' \
    -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'

HTTP/2 200
content-type: application/json
access-control-allow-origin: *
access-control-allow-headers: authorization, x-client-info, apikey, content-type, x-brain-key, accept, mcp-session-id
access-control-allow-methods: GET, POST, OPTIONS, DELETE
... (Cloudflare/Supabase edge headers elided)

{"jsonrpc":"2.0","error":{"code":-32001,"message":"Unauthorized: missing or invalid authentication. Provide both Authorization Bearer token and x-brain-key (header or query param)."},"id":1}
```

Status 200, JSON-RPC `error.code: -32001`, request `id: 1` echoed back. One note on §3's scope: my downstream's `UNAUTHORIZED_MESSAGE` string ("Provide both Authorization Bearer token and x-brain-key…") differs from the upstream PR's shorter `"Unauthorized: missing or invalid authentication."` — the **JSON-RPC structure** is the same (`jsonrpc`, `error.code`, `error.message`, `id`), not a byte-identical message string.

### 4. Pre-fix behavior — wire evidence now captured (above)

The artifact section at the top of this comment is the concrete pre-fix wire trace that was missing from the initial draft. The BEFORE-side code in this PR's own diff (hunk `@@ -381,7 +458,14 @@`, in the BEFORE side) reads:

```ts
return c.json({ error: "Invalid or missing access key" }, 401, corsHeaders);
```

Hono's `c.json(body, status, headers)` — a bare HTTP 401 with a plain JSON body containing only an `error` string, no `jsonrpc`, no `error.code`, no `id` correlation. The captured Codex log above shows exactly what happens when `rmcp`'s deserializer receives that response during `initialize`.

### 5. A note on Open Brain's deployment relationship to OB1

Open Brain is a parallel personal-use deployment running the same Supabase Edge Function pattern (`open-brain-mcp`). It is not a fork of OB1's repository — it was authored independently. The fix in this PR is a port upstream of the same change I deployed to my own instance on 2026-04-26 (commit `1e9facd`). Wire behavior on the post-fix path should be structurally identical between the two deployments once this lands.

### 6. Security trade and operational impact

The auth predicate, as you noted, is byte-identical. No bypass is introduced. The realistic operational cost of the status-code change:

- Any uptime/synthetic check that asserts "401 on bad key" needs to assert on JSON-RPC `error.code === -32001` instead. Small but non-zero migration cost.
- Rate-limit or WAF rules keyed on `4xx` from the `/mcp` endpoint will no longer fire on auth-fail traffic. Whether that's bad depends on the deployment's threat model.
- Logging/observability pipelines that aggregate by status code lose this signal; aggregation by JSON-RPC error code recovers it.

### 7. Maintainer call

The status-code convention itself is Nate's to set. The empirical question for the alternative (`HTTP 401 + JSON-RPC envelope in the body`) is whether strict MCP hosts still tear the transport down on a 4xx *even when* the body parses as a JSON-RPC error — `rmcp`'s deserializer panic above appears to fire before it checks status code (inferred from log order — I haven't read rmcp source to confirm which check runs first), but I haven't tested the case where the body is valid JSON-RPC and status is 401 to know which order those checks run. That's testable with the same mock server approach above (change the status line, keep the body valid). Options on the table:

- 200 + JSON-RPC envelope (what this PR does; confirmed working with Codex CLI).
- 401 + JSON-RPC envelope (untested against strict hosts; may still trigger transport teardown, may not).
- 401 with the current opaque error body (status quo upstream; triggers transport teardown per artifact above).

---

### Appendix: Full reproduction

These steps reproduce the pre-fix vs post-fix comparison in approximately 5 minutes. All components are available without deploying anything.

#### Mock server (Python 3, no dependencies)

```python
#!/usr/bin/env python3
"""
Mock HTTP server v2 — mimics the PRE-FIX wire shape of open-brain-mcp
for POSTs, while gracefully returning 404 for unknown GETs
such as OAuth discovery probes (/.well-known/oauth-authorization-server).

Behavior:
  - POST anything  -> HTTP 401  + body {"error":"Invalid or missing access key"}
  - GET  anything  -> HTTP 404  + body {"error":"not found"}
  - OPTIONS        -> HTTP 200 with permissive CORS

Logs every request (method, path, response status) to stderr.

Usage: python3 mock-prefix-401-v2.py [port]
Default port: 8766
"""

import sys
import json
from http.server import HTTPServer, BaseHTTPRequestHandler

PORT = int(sys.argv[1]) if len(sys.argv) > 1 else 8766


class PreFixV2Handler(BaseHTTPRequestHandler):
    def _respond(self, status, payload):
        body = json.dumps(payload).encode()
        self.send_response(status)
        self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(body)))
        self.send_header("Access-Control-Allow-Origin", "*")
        self.send_header(
            "Access-Control-Allow-Headers",
            "content-type, x-brain-key, authorization, mcp-session-id, mcp-protocol-version, accept",
        )
        self.send_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS, DELETE")
        self.end_headers()
        self.wfile.write(body)
        sys.stderr.write(f"[mock-v2] {self.command} {self.path} -> {status}\n")
        sys.stderr.flush()

    def do_POST(self):
        content_length = int(self.headers.get("Content-Length", 0))
        _ = self.rfile.read(content_length) if content_length else b""
        self._respond(401, {"error": "Invalid or missing access key"})

    def do_GET(self):
        self._respond(404, {"error": "not found"})

    def do_DELETE(self):
        self._respond(404, {"error": "not found"})

    def do_OPTIONS(self):
        self.send_response(200)
        self.send_header("Access-Control-Allow-Origin", "*")
        self.send_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS, DELETE")
        self.send_header(
            "Access-Control-Allow-Headers",
            "content-type, x-brain-key, authorization, mcp-session-id, mcp-protocol-version, accept",
        )
        self.end_headers()
        sys.stderr.write(f"[mock-v2] OPTIONS {self.path} -> 200\n")
        sys.stderr.flush()

    def log_message(self, format, *args):
        return  # suppress default access log


if __name__ == "__main__":
    server = HTTPServer(("127.0.0.1", PORT), PreFixV2Handler)
    print(f"PRE-FIX mock v2 listening on http://127.0.0.1:{PORT}", flush=True)
    server.serve_forever()
```

Save this as `mock-prefix-401-v2.py`. Start it:

```bash
python3 mock-prefix-401-v2.py 8766 &
```

#### Step 1 — Verify mock wire shape

```bash
# Should return HTTP/1.0 401 + {"error":"Invalid or missing access key"}
curl -i -X POST http://localhost:8766/mcp -d '{}'

# Should return HTTP/1.0 404 + {"error":"not found"}
curl -i http://localhost:8766/.well-known/oauth-authorization-server
```

#### Step 2 — Codex CLI repro

Version: `0.125.0` (OpenAI Codex research preview)

Register the mock:
```bash
codex mcp add mock-prefix-401 --transport http-client http://localhost:8766/mcp
```

This writes to `~/.codex/config.toml` under `[mcp_servers.mock-prefix-401]`.

Trigger the error:
```bash
codex -p "Use the mock-prefix-401 MCP server to call tools/list and report what you get."
```

Expected error fragment (from stderr, appears before the session header):
```
ERROR rmcp::transport::worker: worker quit with fatal: Transport channel closed,
  when Deserialize(Error("data did not match any variant of untagged enum
  JsonRpcMessage", line: 0, column: 0))
```

Expected session output (approximately 18 seconds later):
```
ERROR codex_core::tools::router: error=resources/list failed: failed to get client:
  MCP startup failed: handshaking with MCP server failed: Send message error
  Transport [...rmcp...] error: Deserialize error: data did not match any variant
  of untagged enum JsonRpcMessage, when send initialize request
mcp: mock-prefix-401/list_mcp_resources started
mcp: mock-prefix-401/list_mcp_resources (failed)
```

The mock server log (stderr) should show only:
```
[mock-v2] POST /mcp -> 401
```

Codex sends exactly one request (the initialize POST) before crashing the transport. No OAuth discovery probes.

#### Step 3 — Claude Code repro

Version: `2.1.144` (Claude Code)

```bash
mkdir -p /tmp/claude-mcp-test && cd /tmp/claude-mcp-test
claude mcp add mock-prefix-401-v2 --transport http http://localhost:8766/mcp
# This writes to ~/.claude.json under the project path /tmp/claude-mcp-test
```

Trigger:
```bash
cd /tmp/claude-mcp-test
claude -p "what mcp tools failed to connect" 2>&1
```

Expected user-visible output:
```
Failed to connect:
- `mock-prefix-401-v2` (HTTP at `http://localhost:8766/mcp`)
```

The mock server log (stderr) for this run shows a multi-step OAuth-first handshake:
```
[mock-v2] POST /mcp -> 401
[mock-v2] GET /.well-known/oauth-authorization-server -> 404
[mock-v2] GET /.well-known/openid-configuration -> 404
[mock-v2] GET /.well-known/oauth-protected-resource/mcp -> 404
[mock-v2] GET /.well-known/oauth-protected-resource -> 404
[mock-v2] POST /register -> 401
```

The `mock-prefix-401-v2` server does not appear in the session's MCP tool inventory. Session continues normally without it; no error is shown unless explicitly asked.

#### How to interpret this

Both clients fail when the endpoint returns a bare HTTP 401, but via different mechanisms:

- **Codex CLI** uses `rmcp`'s Streamable HTTP transport, which attempts `initialize` immediately (one POST). The bare 401 body is not parseable as `JsonRpcMessage`, so the transport worker crashes (deserialize panic) before any MCP method executes. The error is surfaced as `MCP startup failed: handshaking... Deserialize error` — correct semantics, cryptic message, nothing actionable for the user.

- **Claude Code** uses an OAuth-first handshake: initial POST, then multiple OAuth discovery probes (`.well-known/*`), then dynamic client registration (`POST /register`). All paths return 401 or 404. After exhausting OAuth, the server is silently dropped. The session continues without error — the user only learns the server failed if they ask explicitly.

The PR #243 fix (returning `HTTP 200 + {"jsonrpc":"2.0","error":{"code":-32001,...}}`) addresses both clients:
- For Codex: the 200 allows `initialize` to proceed; `rmcp` receives a valid JSON-RPC response body and can surface the `-32001` error cleanly instead of crashing.
- For Claude Code: the 200 satisfies the initial POST check, allowing the transport to progress past the OAuth-discovery phase and surface a structured JSON-RPC error rather than silently dropping the server.

---

Thanks again for the review.

— Scott (`txcfi-scott`)

---

alanshurafa:
This is exactly the evidence I asked for — thanks for the depth here. The Codex CLI `rmcp` deserialize teardown on a bare 401 is now concretely documented, and the Claude Code OAuth-exhaustion path (silent drop, server marked failed) is a useful second data point.

I haven't re-run the repro myself, but it's fully scripted in your appendix, so the maintainer can reproduce both cases in a few minutes.

One open thread, from your own §7: the 401 + valid JSON-RPC envelope variant is still untested. If a 4xx with a parseable JSON-RPC body doesn't trip the transport teardown, that's the option that keeps the conventional status code and the structured error. Worth knowing before the maintainer picks between your three options — not blocking, just the piece that would make the call fully informed.

From my side this moves to ready-for-maintainer. The 200-vs-401 convention is the maintainer's to set, but it's now a well-supported decision either way.

— Alan

justfinethanku (approved):
Maintainer validation complete. I checked the one-file server diff, confirmed the auth predicate is unchanged, ran , and smoke-tested the local wire behavior: wrong-key POST now returns a JSON-RPC 2.0 error envelope with code -32001 and echoed id, malformed JSON falls back to id:null, OPTIONS remains unchanged, and a valid initialize request still reaches StreamableHTTPTransport. Approving this compatibility fix.

PR #252, opened 2026-04-28
recipes: local-ollama-embeddings — upsert, reembed, content_fingerprint

snapsynapse:
## Summary

Adds three production features to `embed-local.py` that make the local-embeddings recipe usable as a repeatable seed pipeline:

- **`--upsert`** — uses `ON CONFLICT (content_fingerprint)` with `resolution=merge-duplicates` via the PostgREST `Prefer` header. Requires a `UNIQUE` constraint on `content_fingerprint`; no-ops gracefully if the constraint is absent (falls back to plain insert).
- **`--reembed`** + **`--reembed-limit N`** — fetches all existing rows (`id` + `content`), re-generates embeddings via Ollama, and PATCHes the `embedding` column. Useful after switching models (e.g. nomic-embed-text → mxbai-embed-large). Does not touch `content` or `metadata`.
- **`content_fingerprint` passthrough** — JSONL inputs with a `content_fingerprint` key now forward the value to the insert/upsert call (was silently dropped before).

### New helpers

- `_supa_headers(prefer)` — extracted from inline dicts to remove duplication
- `update_embedding(row_id, embedding)` — PATCH for reembed path
- `fetch_all_rows()` — GET `id,content` ordered by `created_at`

### Input collection change

When `--reembed` is active, the script skips stdin/file/arg input collection entirely (no `--file` or piped input required).

## Test plan

- [ ] `python embed-local.py --file thoughts.jsonl --upsert` — confirm ON CONFLICT merge on second run
- [ ] `python embed-local.py --reembed --dry-run` — confirm rows fetched, embeddings generated, no DB writes
- [ ] `python embed-local.py --reembed --reembed-limit 5` — confirm only 5 rows patched
- [ ] `python embed-local.py --file plain.txt` — confirm unchanged plain-insert path still works
- [ ] `python embed-local.py` (no args, no stdin) — confirm still exits with clear error


alanshurafa:
This branch conflicts with current `main`. Recommend author refresh — a rebase should clear it. One thing worth a README note on the rebase pass: the `--upsert` path assumes a `UNIQUE(content_fingerprint)` constraint, which not every brain has, so calling that out would help adopters.

PR #270, opened 2026-05-06
docs: rewrite FAQ for two-header auth pattern (v7)

brmarketing:
## Summary

- URL-param auth (`?key=`) was removed in `open-brain-mcp` v7 (2026-05-04) for security — keys in URLs leak to server logs, browser history, and HTTP referer headers
- Rewrites the setup section with a two-header pattern: `Authorization: Bearer <SUPABASE_ANON_KEY>` + `x-brain-key: <OPEN_BRAIN_ACCESS_KEY>`
- Adds "Why two headers?" explanation (Supabase gateway layer vs app auth layer)
- Adds concrete config examples for Claude Desktop, ChatGPT custom GPT, and Cursor/Claude Code
- Adds deprecation note for users still on the old `?key=` config
- Updates Quick Reference item 3 to match

## Test plan

- [ ] Verify Claude Desktop example JSON parses correctly
- [ ] Verify ChatGPT Action schema block is valid JSON
- [ ] Verify `.mcp.json` block is valid JSON
- [ ] Confirm no remaining `?key=` references in the file

### DIFF
--- a/docs/03-faq.md
+++ b/docs/03-faq.md
@@ -4,15 +4,68 @@
 
 ### "Claude Desktop / ChatGPT gives me an auth error but Claude Code works fine"
 
-This is the single most common issue. The tell is right in the pattern: Claude Code can send custom headers, but Claude Desktop, Claude Web, and ChatGPT can't.
+This is the single most common setup issue. Every Open Brain connection requires **two headers** — one for Supabase's gateway, one for the Open Brain itself. Missing either one returns a 401 with no useful explanation.
 
-The fix: use the MCP Connection URL with the key embedded as a query parameter (`?key=your-access-key`), not as a custom header. Your URL should look like:
+#### Why two headers?
 
-```text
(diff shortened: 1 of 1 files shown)


alanshurafa:
Thanks for the contribution, and welcome. The two-header explanation and the per-client config blocks (Claude Desktop, ChatGPT, Cursor) are useful additions.

One thing needs reconciling before this is ready. The PR states that URL-parameter auth (`?key=`) "was removed in v7" and that configs still using it "will receive a 401 after the v7 deploy." On the current `main`, `server/index.ts` (line 524) still reads:

    const provided = c.req.header("x-brain-key") || new URL(c.req.url).searchParams.get("key");

with the comment "Accept access key via header OR URL query parameter." So `?key=` is still accepted in the repo as it stands, and the FAQ would tell users their working setup is deprecated when it is not.

Could you point at where the v7 `?key=` removal happens in this repo, or soften the migration note so it does not pre-announce a change that has not merged? Recommend author refresh once the doc lines up with the server.

Issue #279, opened 2026-05-09
[Bug] @natebjones/ob1-agent-memory v0.1.1 — plugin loads as memory-slot owner but neither auto-invokes nor exposes its 7 tools to agents in OpenClaw 2026.5.7


MicScalise:
### Which contribution?

`integrations/openclaw-agent-memory` (the `@natebjones/ob1-agent-memory` plugin, v0.1.1)

### What happened?

The plugin installs and loads cleanly under OpenClaw 2026.5.7, owns the exclusive `memory` slot, and the gateway reports it as active — but **neither auto-invocation nor agent-callable tools work**, so the seven OB1 Agent Memory tools (`openbrain_recall`, `openbrain_writeback`, etc.) are unreachable from any agent.

Concretely:

- `openclaw plugins doctor` → "No plugin issues detected."
- Gateway log on startup: `http server listening (2 plugins: nbj-ob1-agent-memory, telegram; 2.5s)` ✓
- `plugins.slots.memory: nbj-ob1-agent-memory` is set in `openclaw.json`
- After dispatching any agent through the gateway:
  - The agent never sees `openbrain_recall` etc. in its tool list. Asking "list any tools whose name starts with `openbrain_`" returns the unrelated leftover MCP tool surface (`openbrain__capture_thought` etc., double-underscore namespacing).
  - Calling `openbrain_recall` directly returns: `Error: Tool 'openbrain_recall' not found in available tools.`
  - The `agent_memories`, `agent_memory_recall_traces`, and `agent_memory_audit_events` tables in OpenBrain stay at **0 rows**, and the Edge Function logs receive **no traffic** from the agent run, even though the agent's SOUL/identity loads and the turn completes successfully.

The Edge Function backend itself is healthy — direct `curl` against the configured endpoint returns expected responses (`/health` → 200, `/recall` and `/writeback` validate input shape and reject correctly), so the issue is between the plugin and the OpenClaw runtime, not in the API.

### What did you expect?

Per the v0.1.1 release notes ("a clean OpenClaw profile should install the plugin from ClawHub and list all seven OB1 Agent Memory tools"), and per the `kind: memory` plugin pattern other memory plugins use in OpenClaw (e.g. bundled `memory-core`, `@openclaw/memory-lancedb`), one of:

1. The seven `openbrain_*` tools appear in the agent's available tool list and the agent can call them directly, **or**
[...]

PR #280, opened 2026-05-09
feat(integrations): add hermes-agent-memory native provider for OB1

MicScalise:
## Summary

Adds `integrations/hermes-agent-memory/` — a native Hermes Agent `MemoryProvider` that connects Hermes runtimes to the OB1 v1 Agent Memory contract. Same backend as the OpenClaw plugin, so OpenClaw agents and Hermes agents share **one** governed memory.

This is a companion to [`integrations/openclaw-agent-memory/`](https://github.com/NateBJones-Projects/OB1/tree/main/integrations/openclaw-agent-memory) — same governance, different runtime.

## What it does

- **Auto-recall before each turn** via `prefetch()`, with background warm-up via `queue_prefetch()` (90s TTL cache)
- **Auto-writeback after each turn** via `sync_turn()` — payload structured as OB1 `outputs[]`
- **Session-end finding extraction** via `on_session_end()` — heuristic mapping into `decisions / lessons / constraints / next_steps / unresolved_questions / failures`
- **Pre-compression preservation** via `on_pre_compress()` — extracts findings before compression discards messages
- **Seven explicit tools**: `ob1_recall`, `ob1_writeback`, `ob1_search`, `ob1_report_usage`, `ob1_list_review_queue`, `ob1_review_memory`, `ob1_get_recall_trace`
- **x-brain-key auth** (matches OB1 v1 contract — not `Authorization: Bearer`)
- **Per-turn model + provider attribution** — reads `on_turn_start` kwargs, falls back to `~/.hermes/config.yaml` (real `yaml.safe_load`, not line-scanning)
- **Subagent / cron / flush guards** — disables writes for non-primary contexts to prevent corrupting the parent's `task_id`

## Why this exists

OpenClaw and Hermes are sibling agent runtimes. Without a Hermes-native provider, Hermes agents can't participate in the same governed memory the OpenClaw plugin uses — defeating the cross-runtime "shared brain" promise. This plugin closes that gap with no extra infrastructure (talks to the same `agent-memory-api` Edge Function).

## Tests

75 pytest tests, network-mocked at `urllib.request.urlopen`, run offline in <1 second:

\`\`\`bash
cd integrations/hermes-agent-memory/plugin
pytest tests/
# ===== 75 passed in 0.44s =====
\`\`\`

Coverage: pure helpers, OB1 v1 schema shape on the wire (recall + writeback + report_usage), client error handling, lifecycle (active/inactive transitions, subagent guard, save_config sanitization), prefetch + queue_prefetch caching (cold/warm/stale/burst-dedup/shutdown), sync_turn (skip-trivial, skip-short, ob1-context stripping), session-end finding extraction, pre-compress, and all seven tool handlers.

## Verified live

End-to-end against a self-hosted Supabase + `agent-memory-api` Edge Function deployment. Real Hermes turns auto-recall, auto-write back with correct `runtime_name=hermes`, `model=anthropic/claude-opus-4.6`, `provider=anthropic`, and `task_id` linkage. Agent self-organizes around the `system_prompt_block` and calls `ob1_writeback` explicitly for decision/lesson categorization.

## Test plan

- [ ] Tests pass: \`cd integrations/hermes-agent-memory/plugin && pytest tests/\`
- [ ] Install: copy \`plugin/__init__.py\` and \`plugin/plugin.yaml\` to \`~/.hermes/plugins/ob1/\`
- [ ] Configure: \`~/.hermes/ob1.json\` with endpoint + workspace, \`OPENBRAIN_KEY\` in env
- [ ] Activate: \`hermes config set memory.provider ob1\`
- [ ] Drive a turn: \`hermes -z "Decision: ship v0.1.0."\` and verify \`agent_memories\` row lands

See \`integrations/hermes-agent-memory/README.md\` for the full setup walkthrough and \`CHANGELOG.md\` for v0.1.0 release notes.

## Related

Filed alongside #279 (NBJ OpenClaw plugin missing memory-host hooks). This PR's Hermes provider is unrelated to that issue — it's a parallel native implementation for the Hermes runtime side.


alanshurafa:
Thanks for the contribution. This is a substantial, well-organized integration — a Hermes `MemoryProvider` with a CHANGELOG, tests (`test_ob1_provider.py`), and clear docs. It is the Hermes counterpart to the OpenClaw agent-memory plugin: same OB1 Agent Memory backend, different runtime.

Because it sits in the Agent Memory direction, a few things for the maintainer rather than a code fix. It ties to the same direction question as the OpenClaw cluster (#278 / #281 / #283 / #309) — whatever governance model OB1 settles on there should apply consistently here. Worth checking whether `prefetch()` recall carries the same `include_unconfirmed` default question raised on #283. And the integration ships its own `LICENSE` file in the subfolder — worth a quick license-compatibility check against the repo's FSL-1.1-MIT.

Recommend maintainer review, alongside the OpenClaw cluster direction decision.

MicScalise:
Thanks Alan — all three points addressed. New commit `ff4cce0` lands on the PR branch.

**1. LICENSE compatibility (FSL-1.1-MIT)**
You're right — shipping a MIT LICENSE inside an FSL-1.1-MIT repo created a real nesting conflict. Removed `integrations/hermes-agent-memory/LICENSE`; the directory now inherits the repo's root LICENSE.md, matching the convention used by `integrations/openclaw-agent-memory/` which has no subfolder LICENSE.

**2. `prefetch()` recall `include_unconfirmed` default**
Same question as #283 — confirmed. There was a real inconsistency in our code:
- `plugin/__init__.py:96` DEFAULTS dict: `include_unconfirmed_recall: True`
- `plugin/__init__.py:159` config-parse fallback: `False`
- `plugin/__init__.py:944` + `:1292` (both call sites): use the config value
- The Phase-8 final report explicitly says we landed on `True` because recall returned 0 with `False` (pending memories — which is the default for governed writes per `require_review_by_default=True` — were filtered out before reaching the caller).

Flipped line 159 fallback to `True` so the resolved default is consistent with design intent. Edge Function's `scoreMemory` still ranks pending memories lower so they don't dominate confirmed results; they just appear in recall.

**3. Governance direction across the cluster**
Aligned with #283's pattern (governance-discipline on writes via `require_review_by_default=true`; permissive recall via `include_unconfirmed_recall=true`). Happy to update this PR to match whatever the maintainer settles on for the OpenClaw cluster — this Hermes-OB1 integration is the same OB1 backend behind a different runtime, so the resolved governance defaults apply identically. Pinging when the cluster direction lands.

alanshurafa:
Thanks for the quick turnaround — all three points look addressed.

Removing the nested LICENSE is the right call; inheriting the repo root and matching the `openclaw-agent-memory` convention is consistent. On `include_unconfirmed_recall`, good catch on the DEFAULTS-vs-fallback mismatch — that was a real latent inconsistency, and resolving it to `True` lines up with the recall-returns-zero behavior you documented. On governance, agreed it stays the maintainer's call for the whole OpenClaw/Hermes cluster; aligning with #283's pattern is a sensible holding position until that direction lands.

I haven't re-read the `ff4cce0` diff line by line, but the three changes as described are all sound. Recommend maintainer review alongside the cluster direction decision.

— Alan

MicScalise:
Thanks Alan — appreciate the second look. Happy to rebase or adjust if anything shifts when the cluster direction lands.

justfinethanku (approved):
Maintainer review: credit/badge added, stale LICENSE README reference removed, nested license concern is resolved, and fresh lint passes after updating against current main. This is isolated as a Hermes runtime provider for the same governed OB1 Agent Memory backend.

PR #289, opened 2026-05-13
[integrations] Support direct OpenAI API provider

DavidSlayback:
## Summary

- Let the MCP server, Open Brain REST gateway, and Agent Memory API call OpenAI directly.
- Keep OpenRouter as the default and support generic OpenAI-compatible endpoints.
- Update setup docs and fix the fork PR-gate syntax without changing contribution policy.

## Verification

- Deno checks, Markdown lint, metadata validation, diff checks, and secret scans passed.
- A live Supabase/Open Brain smoke test was not run because no project is configured here.

## Next

Address the maintainer's requested clarification, then run or obtain a live smoke test and merge after approval.


alanshurafa:
Thanks for the contribution, and welcome. The core feature — a backwards-compatible direct OpenAI provider path alongside OpenRouter — is a sensible addition, and keeping OpenRouter the default means existing deployments are unaffected.

The main issue is scope. This PR also bundles a change to `.github/workflows/ob1-gate.yml` (moving inline `${{ }}` interpolation into `env:` blocks — a genuine script-injection hardening, but unrelated to the OpenAI provider) and a binary `docs/open-brain-credential-tracker.xlsx` update. The workflow change is worth doing on its own merits, but it is a separate concern, it sits outside what I can review as a community reviewer, and it overlaps #308 (which renames that same file).

Recommend author refresh — splitting the workflow-file change into its own PR would keep #289 scoped to the provider feature plus its docs, make each piece easier to review, and avoid a collision with #308. The binary `.xlsx` is also worth confirming the maintainer wants it tracked in-repo.

PR #290, opened 2026-05-13
fix: add schema_version to writeback method (closes OB1 writeback failure)

cosmo-sage:
## Bug Summary

The `nbj-ob1-agent-memory` plugin v0.1.1 `writeback()` method silently fails because it doesn't send `schema_version` in the request body.

The agent-memory-api server expects `schema_version: "openbrain.openclaw.writeback.v1"` as a required field. Without it, writeback calls fail validation.

## The Fix

Added `schema_version: "openbrain.openclaw.writeback.v1"` to the writeback body object, mirroring the pattern already used in the `recall()` method.

## Impact

- Affects all 3 nodes: Slate, Clay, Doctor (all running v0.1.1)
- `recall()` works because schema_version is included there
- `writeback()` broken — session-end capture fails silently
- Workaround: direct REST calls with correct schema string work fine

## Test

```bash
curl -X POST "https://gjclaovlofzdmawbbaly.supabase.co/functions/v1/agent-memory-api/writeback" \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <access_key>" \
  -d '{"schema_version":"openbrain.openclaw.writeback.v1","workspace_id":"test","memory_payload":{},"provenance":{}}'
```

### DIFF
--- a/integrations/openclaw-agent-memory/plugin/src/client.ts
+++ b/integrations/openclaw-agent-memory/plugin/src/client.ts
@@ -62,6 +62,7 @@ export class AgentMemoryClient {
       body: {
         workspace_id: this.config.workspaceId,
         project_id: this.config.projectId ?? null,
+        schema_version: "openbrain.openclaw.writeback.v1",
         ...input,
         provenance: {
           default_status: "generated",


jonbjornn:
I would suggest putting `schema_version` after the `...input` parameters from the model. In case the model decides to supply this value and supplies an incompatible string like "v1.0" or "2026.x.x". As my testing showed would sometimes happen. Simply changing the order prevents this issue:
```
    schema_version: "openbrain.openclaw.writeback.v1",
    ...input,
    provenance: {
```
vs
```
    ...input,
    schema_version: "openbrain.openclaw.writeback.v1",
    provenance: {
```

alanshurafa:
Thanks for the contribution, and for tracking down the writeback failure. The fix is on the right track.

Worth folding in before this moves forward: @jonbjornn's comment above is a valid catch. As written, `schema_version` sits before the `...input` spread, so if the model supplies its own `schema_version` the model's value would override the hardcoded `"openbrain.openclaw.writeback.v1"`. Moving the line to just after `...input` makes the intended value always win:

```
...input,
schema_version: "openbrain.openclaw.writeback.v1",
provenance: {
```

That is a one-line reorder. Recommend author refresh with that change before this goes to a maintainer.

PR #305, opened 2026-05-15
[schemas] Make enhanced-thoughts self-contained for status columns

lucifer:
## Contribution Type

- [x] Schema (`/schemas`)

## What does this do?

Fixes a hidden cross-schema dependency in `schemas/enhanced-thoughts/schema.sql`. The `upsert_thought` RPC defined in that file writes to `status` and `status_updated_at`, but those columns are only added by `schemas/workflow-status/migration.sql`. `docs/01-getting-started.md` only instructs users to apply `enhanced-thoughts`, so following the getting-started path produces a broken RPC and `capture_thought` fails with:

> column "status" of relation "thoughts" does not exist

This PR adds idempotent `ADD COLUMN IF NOT EXISTS` for `status` / `status_updated_at` (and the matching partial index) directly in `enhanced-thoughts/schema.sql`, so the RPC's dependencies live in the same file. Both schemas use `IF NOT EXISTS`, so users who already ran `workflow-status` are unaffected.

## Requirements

None — pure SQL change in an existing schema file.

## Checklist

- [ ] My contribution has a `README.md` with prerequisites, step-by-step instructions, and expected outcome — N/A, edit to existing schema
- [ ] My `metadata.json` has all required fields — N/A, no metadata change
- [ ] If my contribution depends on a skill or primitive, I declared it in metadata.json and linked it in the README — N/A
- [x] I tested this on my own Open Brain instance (ran the equivalent ALTERs against my Supabase; `capture_thought` now succeeds)


alanshurafa:
Thanks for the contribution. This closes a real gap: `enhanced-thoughts` ships the `upsert_thought` RPC that writes `status` and `status_updated_at`, but those columns were only declared in `schemas/workflow-status`. Installing `enhanced-thoughts` alone would hit a missing-column error.

Checked the column definitions against `schemas/workflow-status/migration.sql`: the `status` / `status_updated_at` types and defaults and the `idx_thoughts_status` partial index are identical in both files, so applying either or both in any order leaves the same result. The inline comment documenting the dual definition is accurate. Recommend maintainer review.

justfinethanku (approved):
Maintainer review complete. This is the smallest safe fix for the missing status/status_updated_at dependency in enhanced-thoughts, and Alan already verified the definitions match workflow-status.

PR #283, opened 2026-05-10
feat(openclaw-agent-memory): register OB1 as active-memory capability for governed recall and writeback (#279)

MicScalise:
Closes the architecture gap left by PR #281. Companion to #280.

## What this changes

PR #281 added two additive memory hooks (`registerMemoryPromptSupplement`, `registerMemoryCorpusSupplement`) so OB1 results contribute alongside whatever active memory plugin happens to be selected. That was the safe minimum. It did **not** make OB1 _be_ the active memory plugin — agents still couldn't reach OB1 from `memory_search`, and when `plugins.slots.memory = "nbj-ob1-agent-memory"` was set the slot stayed empty (\`openclaw doctor\` reported "No active memory plugin is registered for the current config").

This PR fills the slot. Inside `register(api)`:

- **`api.registerMemoryCapability({ promptBuilder, runtime })`** — emits an OB1 discipline section into the active memory prompt and provides a real `MemoryPluginRuntime` that produces per-agent search managers.
- **`OB1MemorySearchManager`** (`src/search-manager.ts`, new) — implements the surface OpenClaw consumers actually call: `search`, `status`, `probeVectorAvailability`, `close`, `hasIndexedContent`, `warmSession`, `sync`. `search()` wraps `client.recall()` and maps OB1 memories to OpenClaw's chunk shape (`{source, path, startLine, endLine, score, snippet}`) using a synthetic `openbrain://memory/<id>` path. `status()` reports `backend="ob1"` so downstream code that branches on `backend === "qmd"` treats us as a generic plugin-owned backend.
- **Standard tool wrappers** — `memory_search`, `memory_get`, `memory_store` register under the names OpenClaw's host and active-memory plugin look for. The seven existing `openbrain_*` tools stay registered as the advanced governance surface (review queue, recall traces, explicit usage reporting) and remain unchanged.
- **`memory_store` semantics** — typed `memory_type` parameter maps to OB1 governance categories (`decisions`, `lessons`, `constraints`, `outputs`, `failures`, `next_steps`, `unresolved_questions`).
- **`client.ts` schema_version** — every recall and writeback now sends the required `schema_version` literal (`openbrain.openclaw.recall.v1` / `openbrain.openclaw.writeback.v1`). Without it the Edge Function returned `400 Invalid recall payload`. This was a latent bug because no agent had actually invoked the plugin's recall path before this work — pure tool-call paths bypassed the issue.
[...]

PR #309, opened 2026-05-17
[integrations] Add OpenClaw Agent Memory tool schemas

Humestone:
## Summary
- replace broad `Type.Record(Type.String(), Type.Any())` parameters for `openbrain_recall` and `openbrain_writeback` with explicit Typebox schemas
- mirror the Agent Memory API request shapes for recall scope/limits/entities and write-back memory payload/provenance/source references
- rebuild `integrations/openclaw-agent-memory/plugin/dist/index.js` from the updated source

## Why
This keeps the OpenClaw plugin tool surface inspectable and safer for agents without changing the API client behavior. It is intentionally complementary to #278: this PR only describes the tool parameters; #278 can still inject schema versions and improve API error detail.

## Verification
- `npm install --ignore-scripts --omit=peer`
- `npm run build`
- `node --check dist/index.js`
- `git diff --check`

## Notes
- `workspace_id` and `schema_version` remain optional at the tool boundary because the plugin/client path and pending #278 own runtime injection behavior.
- This does not add new tools, change network behavior, or touch live Agent Memory data.

Related to #278, #279.

### DIFF
--- a/integrations/openclaw-agent-memory/plugin/dist/index.js
+++ b/integrations/openclaw-agent-memory/plugin/dist/index.js
@@ -123,6 +123,112 @@ function toolResult(value) {
     details: value
   };
 }
+var nullableString = Type.Union([Type.String(), Type.Literal(null)]);
+var optionalNullableString = Type.Optional(nullableString);
+var stringArrayRecord = Type.Record(Type.String(), Type.Array(Type.String()));
+var channelParameters = Type.Object({
+  kind: optionalNullableString,
+  id: optionalNullableString,
+  thread_id: optionalNullableString
+});
+var runtimeParameters = Type.Object({
+  name: Type.Optional(Type.String()),
+  version: optionalNullableString
+});
+var modelIntentParameters = Type.Object({
+  provider: optionalNullableString,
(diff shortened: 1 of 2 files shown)


alanshurafa:
Thanks for the contribution. The explicit Typebox schemas for `openbrain_recall` and `openbrain_writeback` mirror the Edge Function's expected shapes closely, and keeping `schema_version` optional at the tool boundary is the right call — runtime injection is the client's job. The bounded `limits` (max_items and max_tokens ranges) are a good touch for constraining the model. `dist/index.js` is regenerated, so the built plugin reflects the change.

One coordination note for the maintainer: #309 modifies `src/index.ts`, which #281 and #283 also touch — whichever of the cluster lands first, the others will need a small rebase. Merge ordering across the four-PR OpenClaw cluster (#278, #281, #283, #309) is flagged separately. Recommend maintainer review.

justfinethanku (approved):
Reviewed in the maintainer merge lane. OpenClaw plugin rebuilds cleanly after refresh and schema checks pass; no unresolved review concerns remain.

Issue #262, opened 2026-05-05
[recipes] schema-aware-routing: database-schema.sql fails on any existing OB1 install


AppleCedAR:
### Which contribution?

recipes/schema-aware-routing

### What happened?

Running `database-schema.sql` against a Supabase project that already has the canonical `thoughts` table (from `docs/01-getting-started.md`) fails with:

```
ERROR: 42703: column "domain" does not exist
```

No new tables or columns are created — the Supabase SQL Editor rolls the whole transaction back, leaving the install in a broken state.

**Root cause.** The script wraps the `thoughts` extension in `create table if not exists thoughts (...)` with three extra columns (`domain`, `status`, `source`). Because the canonical `thoughts` table already exists, `create table if not exists` is a no-op and the three extra columns are never added. The first index that references one of them — `create index if not exists idx_thoughts_domain on thoughts(domain)` — then fails because `domain` doesn't exist on the existing table.

Per `CLAUDE.md`, adding columns to `thoughts` is permitted, but the mechanism must be `alter table ... add column if not exists`, not `create table if not exists`.

**Severity:** high — the recipe is unusable out of the box for anyone who has followed the canonical OB1 setup.

**Documentation follow-ups (same root issue, different files):**
- `recipes/schema-aware-routing/README.md` Step 1 has the same SQL embedded — needs the same fix.
- `recipes/schema-aware-routing/SETUP.md` Quick Setup Step 1 — wording should be "extends" rather than "creates" the `thoughts` table.
- `recipes/schema-aware-routing/setup-checklist.md` Step 1 — references the same SQL.

### What did you expect?

The script should idempotently extend the existing canonical `thoughts` table with the three recipe-specific columns and then proceed to create the recipe's other tables, indexes, and grants.

**Proposed fix.** Replace the `create table if not exists thoughts (...)` block in `database-schema.sql` with:

```sql
-- Assumes canonical thoughts table from docs/01-getting-started.md already exists.
-- Extend it with the recipe-specific columns.
alter table thoughts add column if not exists domain text default 'personal';
alter table thoughts add column if not exists status text default 'active';
alter table thoughts add column if not exists source text default 'api';
```

The rest of the file (other four tables, grants, indexes) works as-is and is already idempotent thanks to `create table if not exists` / `create index if not exists`.

### Steps to reproduce

1. Set up an OB1 Supabase project and run the canonical `thoughts` table SQL from `docs/01-getting-started.md`.
[...]

PR #310, opened 2026-05-17
[recipes] Make schema-aware routing SQL idempotent

Humestone:
## Summary
- update the schema-aware-routing README setup SQL so it uses `create table if not exists` for recipe tables
- extend an existing canonical `thoughts` table with `domain`, `status`, and `source` via `alter table ... add column if not exists`
- add supporting indexes for the recipe-specific `thoughts` columns and clarify the SQL is safe after the canonical getting-started setup

## Why
The current README tells users to create `thoughts` from scratch. That fails or conflicts for users who already followed the canonical OB1 getting-started SQL. This keeps the recipe aligned with OB1's shared `thoughts` table contract while preserving the standalone setup path.

## Verification
- reviewed the README SQL block against `docs/01-getting-started.md`
- `rg` check for non-idempotent table creation in the edited block
- `git diff --check`

## Notes
This PR only updates the setup SQL embedded in the README on current `main`; it does not add missing scaffold files or new runtime behavior.

Related to #262.

### DIFF
--- a/recipes/schema-aware-routing/README.md
+++ b/recipes/schema-aware-routing/README.md
@@ -74,27 +74,39 @@ The LLM is prompted to ONLY extract action items when the speaker commits to doi
 ![Step 1](https://img.shields.io/badge/Step_1-Create_Your_Database_Tables-2E86AB?style=for-the-badge)
 
 <details>
-<summary>📋 <strong>SQL: Create all five tables and grant permissions</strong> (click to expand)</summary>
+<summary>📋 <strong>SQL: Extend thoughts, create recipe tables, and grant permissions</strong> (click to expand)</summary>
 
 ```sql
 -- Enable the vector extension for embeddings
 create extension if not exists vector;
 
--- 1. Thoughts table — the raw capture
-create table thoughts (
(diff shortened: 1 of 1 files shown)


alanshurafa:
Thanks for the contribution. Making the recipe SQL idempotent is the right direction — `create table if not exists` plus `alter ... add column if not exists` lets it run cleanly after the canonical OB1 setup.

One thing to resolve before this is ready: the recipe adds `status text default 'active'` to the shared `thoughts` table, but `schemas/workflow-status` already declares `status` on `thoughts` as `text default null` and uses it for kanban-style task/idea state. Since `add column if not exists` skips when the column is already present, both the default and the meaning of `status` would depend on which setup ran first. Could you confirm whether this recipe's `status` is the same concept? If it is, matching `default null` keeps the two consistent; if it is a distinct routing concept, a different column name would avoid the collision. (`domain` and `source` look recipe-specific — no conflict there.)

Minor: the `alter ... add column if not exists` lines for `metadata`, `created_at`, and `updated_at` duplicate columns the `create table` block above already declares — worth trimming for clarity.

Recommend author refresh once the `status` question is settled.

PR #321, opened 2026-05-18
[schemas] Add per-agent identity primitive

### LINKED ISSUE #256: Proposal: per-agent identity primitive for multi-agent Open Brain fleets
## Summary

I am running Open Brain as the shared memory substrate for a small fleet of long-running agents. The single-user `MCP_ACCESS_KEY` model works well for a personal brain, but a multi-agent deployment needs a lightweight way to identify which agent is making a memory request without trusting a prompt-asserted `agent_id`.

This issue proposes a small per-agent identity primitive:

- each agent receives an opaque memory key
- the Open Brain server stores only a SHA-256 hash of that key
- every MCP request can include the raw agent memory key in a header or query param
- the server derives the authenticated `agent_id` from the key lookup
- downstream memory policy can use that derived identity for ownership, revocation, and lane isolation

## Why this matters

In a multi-agent setup, several Claude Code, Codex, scheduled, or background agents may write to the same Open Brain deployment. Without server-side agent identity, deployments either:

- use one shared key for every agent
- trust the model/runtime to self-report `agent_id`
- cannot revoke one agent without rotating shared credentials
- cannot safely enforce private-agent memory lanes

The design goal is not to replace `MCP_ACCESS_KEY`. It is to add an optional second layer for deployments that need agent attribution and isolation.

## Proposed primitive

A small schema recipe, tentatively `schemas/per-agent-identity/`, with:

- `agent_memory_keys` table
- `agent_id`
- `key_hash`
- `label`
- `active`
- `last_used_at`
- optional `revoked_at`
- `lookup_agent_memory_key(p_key_hash)` SECURITY DEFINER RPC

The lookup RPC returns the active `agent_id` for a presented hash and updates `last_used_at`. Revoked or inactive keys return null.

Raw keys are never stored in Open Brain.

## Expected server usage

At request time:

1. Read `x-agent-memory-key` or `agent_key`.
2. Hash it with SHA-256.
3. Call `lookup_agent_memory_key`.
4. Treat the returned `agent_id` as the authenticated identity.
5. Reject requests where a caller-supplied `agent_id` disagrees with the authenticated identity.

The authenticated agent id should be the basis for future governed memory policy such as:

- `private_agent` ownership
- project memory audience checks
- source attribution
- agent revocation

## Out of scope for this issue

This proposal is deliberately narrow. It does not require changing the base `thoughts` table or existing tools.

Out of scope:

- full governance metadata design
- read policy for private/project/shared lanes
- provisioning UX
- edge function patch details
- migration/backfill of existing memories

Those can build on the primitive after the maintainer agrees on the identity shape.

## Compatibility

- Optional for single-user deployments.
- Coexists with the existing `MCP_ACCESS_KEY` model.
- No data backfill required.
- Existing clients can continue to work unchanged.

## Local pilot status

I have this working in a sandbox deployment with multiple agents. The local implementation uses:

- hashed per-agent keys
- server-side lookup
- private-agent lane enforcement
- revocation checks
- startup recall and answer-time retrieval using the derived identity

I also have a local PR draft prepared, but I am opening this issue first to confirm whether the primitive fits upstream direction before submitting code.

jeremylahners:
## Summary

Adds an optional `schemas/per-agent-identity` contribution for multi-agent Open Brain deployments.

This implements the narrow identity primitive proposed in #256:

- stable `openbrain_agents.canonical_agent_id` identities
- hashed-only `agent_memory_keys` rows
- `active`, `revoked_at`, and `last_used_at` key lifecycle fields
- `lookup_agent_memory_key(p_key_hash)` as a `SECURITY DEFINER` RPC
- RLS enabled with service-role policies
- README guidance for server-side hashing, key rotation, and rejecting prompt-supplied identity mismatches

## Why

The current deployment-level `MCP_ACCESS_KEY` model works for single-user setups, but multi-agent deployments need a server-derived identity boundary. A model or runtime payload should not be trusted to self-assert `agent_id` for private-agent memory, revocation, attribution, or future lane policy.

This keeps the primitive optional and compatible with existing OB1 deployments. The deployment-level key remains in place; the per-agent key is an additional identity layer for fleets that need it.

## Design Notes

The PR incorporates the key-rotation concern raised on #256: durable attribution is tied to `canonical_agent_id`, not to the key hash. Rotating a key means inserting a new key row for the same canonical agent, then revoking the old key.

This PR intentionally does not wire the primitive into the Agent Memory API or base MCP server yet. It provides the schema and documented integration pattern first, so maintainers can confirm the identity shape before runtime-specific adoption.

## Validation

Local checks run:

- `jq empty schemas/per-agent-identity/metadata.json`
- local metadata schema-equivalent Node validation
- `npx --yes markdownlint-cli2 --config .github/.markdownlint.jsonc "schemas/per-agent-identity/**/*.md"`
- gate-style folder/required-file/category/sql-safety/secret/local-MCP checks over `schemas/per-agent-identity`
- `git diff --check`

Closes #256.


alanshurafa:
Thanks for the contribution, and welcome. The README, rotation steps, and server integration example are clear and thorough.

Checked `schema.sql` closely since it is identity/auth-adjacent. The `lookup_agent_memory_key` function pins `search_path = public` (the key hardening for `SECURITY DEFINER`), validates the hash length before querying, and is locked down with `REVOKE ALL ... FROM PUBLIC` plus a `service_role`-only grant. RLS is enabled on both new tables, and the schema adds only `openbrain_agents` and `agent_memory_keys` — the base `thoughts` table is untouched, which keeps it within the contribution guard rails.

Since this introduces an identity/auth primitive, whether to adopt it as the canonical multi-agent approach (and how it relates to `schemas/agent-memory/`) is a maintainer decision. Recommend maintainer review.

justfinethanku (approved):
Maintainer review complete. The schema is optional, scoped to new tables and lookup RPCs, and now includes the standard community contribution credit block in the README.

PR #350, opened 2026-06-08
[schemas] Add ingestion jobs lifecycle

alanshurafa:
## Summary

Adds schema support for reviewable ingestion jobs and ingestion items, including lifecycle fields, evidence append helpers, and job recounting.

## Why

The ingestion stack needs a durable review layer between extracted records and trusted OB1 thoughts, with generated/imported memory treated as evidence by default.

## Validation

- Simulated this branch on current `origin/main`
- `markdownlint-cli2 --config .github/.markdownlint.jsonc "**/*.md" "#node_modules"` -> 0 errors
- `check-jsonschema --schemafile .github/metadata.schema.json schemas/ingestion-jobs/metadata.json`
- `git diff --check`

## Notes

This is opened as a draft first so the real upstream OB1 checks can run before marking it ready.


alanshurafa:
Closing in favor of #364 (`schemas/smart-ingest-tables`), which supersedes this PR. Both define the same `ingestion_jobs` / `ingestion_items` tables and helper RPCs, but #364 is the evolved version:

- Column names match the deployed `integrations/smart-ingest` Edge Function (`input_length`; this PR's `input_bytes` does not).
- `source_label` and `sequence` are nullable, matching how the Edge Function and dashboard ingest client actually insert (NOT NULL here makes those inserts fail).
- Drops legacy bigint-signature function overloads so a pre-UUID install can't leave ambiguous RPCs.
- `SECURITY INVOKER` instead of `SECURITY DEFINER`, and a `search_path` that resolves pgcrypto's `digest()` on both older and current Supabase.
- Structured (collision-proof) evidence hashing and normalization of malformed `metadata.evidence`.

Keeping both open would ship two incompatible copies of the same tables. A follow-up PR will reconcile the remaining overlap with the older `schemas/smart-ingest`.

PR #361, opened 2026-06-14
[schemas] Thought work claims for parallel workers

alanshurafa:
## What this adds

A small coordination layer so several workers can process the same pool of thoughts at once without stepping on each other. New folder: `schemas/thought-work-claims/`.

It's the missing piece for any "fan this batch out across N processes" job — enrichment passes, embedding backfills, scoring, consolidation. Run one worker and it's slow; run five and, without coordination, they pick overlapping rows and do the same work twice.

## How it works

- **`public.thought_work_claims`** — one row per `(thought_id, work_type)`. A row means "a worker owns this thought for this kind of work right now."
- **`claim_thoughts(ids, work_type, worker_id, ttl)`** — a worker hands in candidate ids and gets back only the subset it actually won.
- **`release_thought(...)`** — mark one claim `succeeded`/`failed` when done.
- **`release_claims_for_worker(...)`** — on clean shutdown, drop a worker's open claims so others reclaim them immediately.

**Why no row locks are needed.** Exclusivity comes from the `(thought_id, work_type)` primary key plus `INSERT ... ON CONFLICT DO NOTHING`, returning only inserted rows. When two workers submit the same id at the same moment, Postgres lets exactly one insert win and the other conflicts out silently — the loser just sees that id missing from its result. This is the right primitive when workers supply *specific* ids (as opposed to draining an unknown queue, where `FOR UPDATE SKIP LOCKED` is the tool). No advisory locks, no `SELECT ... FOR UPDATE`.

**Crash recovery via TTL.** A worker can die mid-batch and never release. Each claim carries `ttl_expires_at`; `claim_thoughts` reaps inline on every call, deleting still-`claimed` rows whose TTL has passed before it inserts, so a dead worker's thoughts become claimable again. Terminal `succeeded`/`failed` rows are kept as a record and are not reaped.

## Notes for review

- **UUID id contract.** `thought_id UUID REFERENCES public.thoughts(id) ON DELETE CASCADE`, and the RPC arrays are `UUID[]`, matching the canonical Open Brain `thoughts.id`. The README documents the one-line change for a non-canonical `BIGINT` install.
- **Additive and idempotent.** `CREATE TABLE IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`, `CREATE OR REPLACE FUNCTION`. Nothing on `public.thoughts` is altered or dropped. Every `DELETE` is `WHERE`-qualified.
- **Locked down.** RLS on, table and RPCs granted to `service_role` only — no anon/authenticated access. Functions are `SECURITY INVOKER`.
- **README** has Prerequisites, numbered apply steps, Expected outcome, a complete dependency-free Node worker example (config entirely via env vars), and inspect/clear/rollback SQL.

## Testing

Applied the schema to a throwaway Postgres and exercised it: two workers claiming the same five ids produce a 5/0 split with zero overlap; a different `work_type` co-claims the same thoughts; `release_thought` succeeds for the holder and returns false for a non-holder; expiring TTLs lets a third worker reclaim; bulk release frees open claims; an invalid release status raises; re-applying the schema is a no-op; and deleting a thought cascades its claims away.


justfinethanku:
Merged — thanks! Primary-key claims instead of row locks is a clean way to let parallel workers share one brain. Qualified cleanup SQL and service-role-only grants throughout — appreciated.

PR #365, opened 2026-06-15
[schemas] Persistent wiki pages

alanshurafa:
## The regen guard

Wiki generators usually treat a page as disposable: regenerate it and every human edit is gone on the next run. This schema fixes that.

Pages become persistent database objects split into sections, and each section has an owner. A **machine-owned** section (`origin = 'generated'`) regenerates freely. A **human-owned** section (`origin = 'manual'`, or any `locked` section) is protected — a generated write to it does not overwrite the live text. The new draft parks in a pending buffer, and a human promotes it with `wiki_accept_pending`. Same trust model as everywhere else in Open Brain: machine writes propose, a human accepts.

The rule lives in exactly one place — the `wiki_write_section` RPC — so every writer goes through the same guard instead of each generator re-implementing it.

## How it differs from wiki-compiler / wiki-synthesis

Those recipes compile a throwaway markdown artifact each run — no stored page, no section ownership, no history, no protection for a human edit, by design. This schema is the durable storage layer they were missing. A compiler can point its writes at `wiki_write_section` and immediately gain the override guard and revision history. They are complementary, not competing.

## What this PR adds

A new `schemas/wiki-pages/` folder with `schema.sql`, `README.md`, and `metadata.json`. The schema installs:

- `wiki_pages`, `wiki_sections`, `wiki_section_revisions` (append-only history)
- `wiki_upsert_page`, `wiki_write_section` (the guard), `wiki_accept_pending`

## ID contract

UUID throughout, matching OB1's `public.thoughts.id`:

- `wiki_pages.id`, `wiki_sections.id` are UUID
- `wiki_sections.page_id` to `wiki_pages.id` and `wiki_section_revisions.section_id` to `wiki_sections.id` are UUID FKs
- `wiki_sections.evidence_thought_ids` is `UUID[]` referencing `public.thoughts(id)`

`wiki_section_revisions.id` is a bigint identity surrogate — an internal revision sequence, never a thought id. No bigint thought ids and no view triggers; targets public.thoughts(id) UUID directly.

## Safety and sanitization

- RLS on all three tables; granted to `service_role` only, with an explicit `REVOKE ALL ... FROM PUBLIC, anon, authenticated` (defense in depth on projects that blanket-grant new tables).
- RPCs are `SECURITY INVOKER`.
- Idempotent and additive: `CREATE TABLE IF NOT EXISTS`, `CREATE OR REPLACE FUNCTION`, `CREATE INDEX IF NOT EXISTS`, and a guarded `ON CONFLICT DO NOTHING` seed. No destructive table-removal statements; never touches `public.thoughts` columns.
- Seeded with exactly one fictional example page (`getting-started`) and one generated section. No real page content.

## Verification

Live-validated on a throwaway Postgres 18 cluster:

- machine to machine write overwrites in place (`updated`)
- human edit takes ownership (`manual`)
- machine write to a human-owned section parks a draft (`pending`) and leaves the live body untouched
- `locked` generated section also parks (`pending`)
- `wiki_accept_pending` promotes the draft, clears the buffer, keeps the section human-owned, and snapshots a revision (`accepted`); a second call returns `no_pending`
- revision history captures every body change
- re-applying `schema.sql` inserts zero rows (idempotent)

All `ob1-gate-v2` rules were simulated locally and pass.


justfinethanku:
Merged — thank you! This schema is the foundation the whole wiki stack has been waiting on: generators write through one RPC and hand-edited sections survive regeneration. Follow-on wiki PRs can rebase onto main now.

PR #371, opened 2026-06-16
[recipes] Authorship edges — speaker attribution + self author edges

alanshurafa:
## What this adds

A `recipes/authorship-edges/` recipe that brings authorship to spoken and machine-generated capture — the display-name counterpart to the [atomizer](../atomizer/)'s email correspondent resolver.

- **Deterministic classifier** (no LLM) maps a thought's speaker labels to `self`/`other`/`mixed`/`machine`/`unknown`.
- **Self author edges**: links every `self`/`mixed` thought to a configured self-entity via `thought_entities(mention_role = 'author' | 'participant')`.
- **Backfill** (idempotent, keyset-paginated, batched edge lookup) + a **"what I said vs. machine-generated"** SQL query layer.

Safe to merge: additive only (reads/writes the existing `entities` + `thought_entities` tables from the entity-extraction schema, no schema change); no hardcoded identity (self-entity from `OB_SELF_ENTITY_ID` or an upserted, `is_self`-marked name; refuses to bind a non-self entity); `thought_id` passed opaque so it works on UUID or BIGINT brains; no secrets, Node 18+ zero-deps, with a `node test-attribution.mjs` self-test (21 assertions). Named-other → entity resolution is left as a documented extension point.


justfinethanku:
Merged — thank you. Authorship as a real graph edge instead of a hoped-for metadata string is the right call, and the tested classifier plus the is_self guard on self-entity resolution made this easy to accept.

Issue #401, opened 2026-07-06
[recipes] Provenance-aware capture tool (capture_derived_thought) — prerequisite for #397


eazene:
## Why

Blocks #397. The deployed `capture_thought` MCP tool accepts only `content`, so even with the provenance-chains schema fully applied there is no interactive way to write `derived_from` / `derivation_layer` / `derivation_method` / `supersedes`. Verified on a live project: schema ON, read tools ON, but 0 of 158 thoughts carry any provenance because nothing can write it through the connector. Details in NateBJones-Projects/OB1#397 (comment).

## Approach — Option B (separate tool, zero blast radius)

Add a **new** `capture_derived_thought` MCP tool to `recipes/provenance-chains/mcp-tools.ts` (alongside `trace_provenance` / `find_derivatives`). It does NOT modify the canonical `capture_thought`, so the everyday capture hot path used by all clients is untouched. Synthesis skills call the new tool by name when saving a derived artifact.

## Behavior

Mirrors the canonical capture flow (embed + extract metadata + `upsert_thought` + patch), then adds provenance:

- Optional inputs: `derived_from` (UUID strings), `derivation_layer`, `derivation_method`, `supersedes`.
- Validates against the schema's CHECK constraints: keeps only UUID `derived_from` (non-UUIDs like `#412` move to `metadata.provenance.unresolved_refs`); clamps `derivation_layer` to `primary`/`derived` and `derivation_method` to `synthesis`; defaults layer/method to `derived`/`synthesis` when parents are present.
- Writes provenance in two places, both required because `upsert_thought` inserts only `content`/`content_fingerprint`/`metadata` and merges metadata on conflict:
  1. top-level columns via a follow-up `UPDATE` (same patch that writes the embedding),
  2. a `metadata.provenance` mirror that rides through `upsert_thought` and survives re-capture via its `||` merge.
- If no provenance inputs are passed it behaves like a normal capture (layer defaults to `primary` at the DB).

## Verification

1. Plain capture → unchanged row.
2. Synthesis with 4 UUID parents + 1 `#412` → 4 in `derived_from`, `#412` in `unresolved_refs`, layer `derived`, method `synthesis`, mirror present.
3. `trace_provenance` on the new thought returns the 4 parents.
4. Re-capture identical content → mirror survives the duplicate merge.

## Scope

[...]

Issue #397, opened 2026-07-06
[skills] Wire synthesis skills to the provenance-chains capture contract


eazene:
## Follow-up to #392

#392 adds `skills/provenance-chains/`, which defines a capture-time contract for tagging derived artifacts (`derived_from` as flat UUID strings, `derivation_layer`/`derivation_method` CHECK-valid values, and the `metadata.provenance` mirror). That PR is intentionally scoped to the standalone skill and does **not** modify any existing skill.

This issue tracks the going-forward adoption: several skills already capture synthesized artifacts to Open Brain but do not populate the provenance-chains columns, so on a brain with the [provenance-chains schema](../tree/main/schemas/provenance-chains) installed, their outputs land as orphans.

## Skills to wire

Each should apply the `skills/provenance-chains` capture contract when it captures a derived artifact:

- [ ] `panning-for-gold` — captures gold-found synthesis outputs
- [ ] `research-synthesis` — captures the final synthesis/brief
- [ ] `meeting-synthesis` — captures decisions/action-item syntheses
- [ ] `weekly-signal-diff` — captures the weekly digest
- [ ] `competitive-analysis` — captures the competitive brief

## Per-skill change (small, consistent)

When the skill captures a synthesized thought and the provenance schema is present:
- set `derivation_layer: "derived"` and `derivation_method: "synthesis"`
- populate `derived_from` with the **UUID strings** of the source thoughts it combined (drop/resolve non-UUID refs; record dropped ones under `metadata.provenance.unresolved_refs`)
- mirror those into `metadata.provenance` so they survive `upsert_thought` re-captures

Guard it so it's a no-op when the schema isn't installed (the columns won't exist) — e.g. only tag when the source thought IDs are known and the derivation columns are available.

## Why separate PRs

Each skill is an independent, reviewable change with its own blast radius; batching them would turn a clean review into a cross-cutting one. Suggest one PR per skill (or a small grouped PR) referencing this issue.

## Notes

- Contract reference: `skills/provenance-chains/SKILL.md` (the "Write Side: the capture contract" section).
- One-time repair of already-captured artifacts is handled separately by `recipes/provenance-chains/backfill.mjs`; this issue is about keeping **new** captures tagged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0127BucGfQraB5dkZR38MEF1

eazene:
## Prerequisite discovered: the capture tool can't carry provenance

[...]

PR #409, opened 2026-07-06
[recipes] Add capture_derived_thought MCP tool to provenance-chains

eazene:
## What

Adds a third MCP tool — **`capture_derived_thought`** — to `recipes/provenance-chains/mcp-tools.ts`, alongside `trace_provenance` and `find_derivatives`. It captures a synthesized artifact **together with its provenance**, so the derivation graph is populated at capture time instead of only by the one-time `backfill.mjs`.

Prerequisite for #401 / #397. Recipe-only, additive (`2 files changed`): the canonical `capture_thought` is **not modified**, so the everyday capture hot path used by every connected client is untouched (Option B).

## Why

Verified on a live Open Brain: the schema and read tools were installed, but the deployed `capture_thought` accepts only `content`, so nothing could write `derived_from`/`derivation_*` interactively — 0 of 158 thoughts had any provenance. This tool closes that gap.

## How it works

`upsert_thought` writes only `content`/`content_fingerprint`/`metadata` and merges metadata on a fingerprint conflict, so the tool:
1. folds a `provenance` mirror into `metadata` (survives the `||` merge on re-capture);
2. writes the top-level `derived_from`/`derivation_*`/`supersedes` columns via a follow-up `UPDATE` (same patch as the embedding).

Validation matches the schema CHECK constraints: non-UUID `derived_from` refs (e.g. `#412`) are routed to `metadata.provenance.unresolved_refs` instead of poisoning the array; `derivation_layer`/`derivation_method` are constrained to the legal values; layer/method default to `derived`/`synthesis` when parents are present. No provenance inputs → behaves like a normal capture.

## Verified end-to-end (deployed to a live `open-brain-mcp`, then exercised over MCP)

- `tools/list` exposes all 9 tools including `capture_derived_thought`; existing tools unaffected (`verify_jwt` unchanged).
- Capturing with 4 UUID parents + 1 legacy `#412` → row has 4 UUIDs in `derived_from`, `#412` in `metadata.provenance.unresolved_refs`, `derivation_layer='derived'`, `derivation_method='synthesis'`, mirror present.
- `trace_provenance` on the new thought returns all 4 parents.
- Re-capturing identical content dedupes (same id) and the mirror survives.
- Test row cleaned up; parent thoughts untouched.

`deno check` shows no new error class vs. the untouched `server/index.ts` baseline — only the same implicit-any diagnostics the existing `trace_provenance`/`find_derivatives` handlers already have (type-only; Deno Deploy strips types).

Refs #401. Blocks #397.

https://claude.ai/code/session_0127BucGfQraB5dkZR38MEF1

### DIFF
--- a/recipes/provenance-chains/README.md
+++ b/recipes/provenance-chains/README.md
@@ -8,7 +8,7 @@ The [Provenance Chains schema](../../schemas/provenance-chains/) adds four colum
 
 1. **`backfill.mjs`** — One-time script to mark existing derived artifacts (weekly digests, wikis, lint reports, etc.) as `derivation_layer='derived'` and, where the saved artifact exposes thought IDs, populate `derived_from`.
 2. **`eval.mjs`** — Nightly (or on-demand) grader that scores each derived thought on existence / relevance / sufficiency using an LLM, and writes the scores into the thought's metadata so dashboards can surface low-quality chains. Three grader backends: `openrouter` (hosted, default), `stdin` (manual), and `queue` (emit prompts → another worker grades → apply back).
-3. **`mcp-tools.ts`** — Two MCP tool handler snippets (`trace_provenance`, `find_derivatives`) to drop into your `open-brain-mcp` Edge Function. They wrap the SQL helpers and handle redaction of restricted ancestors, cycle tolerance, and node caps.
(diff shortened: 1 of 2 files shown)


justfinethanku:
Merged — thanks! capture_derived_thought finally lets digests and wikis land with their receipts instead of as orphans. The CHECK-aligned enums and the unresolved_refs routing for non-UUID parents were thoughtful touches.

PR #411, opened 2026-07-06
[extensions] Fix PostgREST filter-injection in ILIKE fallback (professional-crm, household-knowledge)

hannahgwilson:
## What it does
`crm_search_contacts` (extensions/professional-crm) and `search_household_items` (extensions/household-knowledge) both build a PostgREST `.or()` filter string by splicing raw user input directly into filter syntax:

```ts
`name.ilike.%${query}%,company.ilike.%${query}%,...`
```

PostgREST's `.or()` filter grammar treats `,` `.` `(` `)` `\` as structural characters (condition separator, `column.operator.value` separator, grouping, escape). A `query` containing any of them reshapes the filter instead of just matching text — not exploitable for data exfiltration since both tools still scope by `user_id`, but a real correctness bug: an ordinary search like `"Smith, John"` or `"Dr. Jane (Referral)"` silently produces wrong or broken results instead of an error.

Each file gets its own small inline `escapePostgrestFilterValue` helper (no shared module — keeps these files independently copy-pasteable, matching this repo's single-file extension convention) that escapes the four reserved characters before interpolating.

## What it requires
No new services/tools — pure logic fix, no schema or dependency changes.

## Testing
- `deno check` passes on both files.
- The identical fix (same vulnerable `.or()` pattern, verbatim) is deployed and running against my own Open Brain instance in a downstream fork of these two extensions. Table/column names differ there (a later refactor unified contacts/vendors into a shared `contacts` table with tag filters) but the vulnerable line itself — and the fix — is unchanged.

### DIFF
--- a/.github/metadata.schema.json
+++ b/.github/metadata.schema.json
@@ -79,6 +79,31 @@
       "items": { "type": "string" },
       "description": "Slugs of primitives this contribution depends on (e.g., [\"rls\", \"shared-mcp\"]). Primarily used by extensions."
     },
+    "contributors": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["name", "contribution"],
+        "properties": {
+          "name": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Contributor's display name."
+          },
+          "github": {
+            "type": "string",
+            "description": "Contributor's GitHub username (without @)."
+          },
+          "contribution": {
+            "type": "string",
+            "minLength": 1,
+            "description": "What this contributor added or changed."
+          }
+        },
+        "additionalProperties": false
+      },
+      "description": "Additional contributors beyond the primary author, each with a note on their specific contribution."
+    },
     "requires_skills": {
       "type": "array",
       "items": { "type": "string" },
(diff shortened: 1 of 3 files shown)


hannahgwilson:
Heads up on the "Metadata valid" gate failure: it's flagging `extensions/professional-crm/metadata.json`'s `contributors` array (added in an earlier merged commit, unrelated to this PR's change) as an unrecognized property against `.github/metadata.schema.json`. This PR doesn't touch that file at all — the gate revalidates metadata.json for any contribution folder touched by a PR, not just changed files, so it's surfacing a pre-existing schema gap here.

I looked at fixing it by adding `contributors` to the schema, but that's a repo-wide file outside a contribution folder, and the gate's own "Scope check" correctly rejects contributor PRs for touching it (plus it validates against `origin/main`'s schema, so an in-PR schema edit wouldn't take effect anyway). That's a maintainer-side fix.

Could a maintainer either update `.github/metadata.schema.json` to allow the `contributors` array (shape: `{name, github, contribution}`), or otherwise reconcile `extensions/professional-crm/metadata.json`? Happy to adjust this PR once that's settled.

PR #425, opened 2026-07-09
[server] Reject non-POST MCP requests with 405 to prevent GET handshake hang

### LINKED ISSUE #424: MCP server hangs forever on GET requests — mcp-remote handshake times out at 60s with no error
## Symptom

Agent (Claude Code / Codex / anything connecting via `mcp-remote`) shows the open-brain MCP server as "running"/"connected", but no tools work. The MCP client log shows a clean startup:

```
Connected to remote server using StreamableHTTPClientTransport
Proxy established successfully between local STDIO and remote StreamableHTTPClientTransport
[Local→Remote] initialize
```

…then nothing, until exactly 60 seconds later the client gives up with `notifications/cancelled` and the transports close with `AbortError`s. No `BOOT_ERROR`, no 5xx, no error server-side. A curl POST smoke test against the endpoint succeeds instantly, which makes this very easy to misdiagnose as a transient cold-start problem.

## Root cause

`server/index.ts` routes **every** HTTP method into `StreamableHTTPTransport.handleRequest()` via `app.all("*")`. For a GET, `@hono/mcp` treats that as a request for the standalone SSE stream and parks the connection on a stream that never emits and never closes — the GET hangs forever.

`mcp-remote` always sends a GET probe to the endpoint (OAuth `WWW-Authenticate` discovery) **before** its initialize POST. That probe hangs server-side, and the initialize POST that follows then hangs non-deterministically (instance-dependent), so the MCP handshake times out at the client.

curl succeeds because it only sends the POST — the failure needs the GET-then-POST sequence that mcp-remote produces, which is why the recommended smoke test passes while the recommended client fails.

## Reproduce (two commands)

```bash
# GET must not hang. Broken = curl times out (exit 28). Fixed = 405 in ~1s.
curl -s -o /dev/null -w "%{http_code} %{time_total}s\n" --max-time 15 \
  'https://<PROJECT>.supabase.co/functions/v1/<FUNCTION>?key=<KEY>'

# POST works either way — which is what makes this bug so misleading:
curl -s -X POST 'https://<PROJECT>.supabase.co/functions/v1/<FUNCTION>?key=<KEY>' \
  -H 'content-type: application/json' -H 'accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":0,"method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{},"clientInfo":{"name":"curl","version":"0"}}}'
```

## Fix

Per the MCP streamable-HTTP spec, a server that doesn't offer a standalone SSE stream (GET) or session termination (DELETE) must answer those methods with **405 Method Not Allowed**. A guard ahead of the transport fixes it:

```ts
if (c.req.method !== "POST") {
  return c.json({ error: "Method not allowed" }, 405, { ...corsHeaders, Allow: "POST, OPTIONS" });
}
```

## Affected files (all `app.all("*")` → `handleRequest` sites)

- `server/index.ts` — the core MCP server
- `integrations/delete-thought-mcp/index.ts`
- `integrations/update-thought-mcp/index.ts`
- `integrations/kubernetes-deployment/index.ts`
- `integrations/enhanced-mcp/index.ts`
- `recipes/edge-function-cost-optimization/examples/after/index.ts` (+ README snippet)
- `recipes/work-operating-model-activation/index.ts`
- `extensions/_template/AGENT_SPEC.md` — the template new extensions copy from

Extensions using `app.post("*")` (meal-planning, professional-crm, job-hunt, home-maintenance, family-calendar, household-knowledge) are **not** affected — Hono 404s other methods before they reach the transport.

Verified on my own deployment (Supabase Edge Functions, project on the current `server/index.ts` pattern): before the guard, every mcp-remote attach timed out at 60s; after deploying the guard, GET returns 405 in ~1s and the full initialize → tools/list → tool-call path works. I have a PR ready with the guard applied to all affected files.

jcpoyser:
Fixes #424.

### What

Adds a `405 Method Not Allowed` guard ahead of `StreamableHTTPTransport.handleRequest()` in the core server, and in the extension template's handler so new extensions don't reinherit the pattern. Per the MCP streamable-HTTP spec, a server that doesn't offer a standalone SSE stream (GET) or session termination (DELETE) must answer those methods with 405.

### Why

Without the guard, a GET request parks forever on a never-emitting, never-closing SSE stream inside `@hono/mcp`. `mcp-remote` — the client the OB1 docs recommend — always sends a GET probe (OAuth `WWW-Authenticate` discovery) before its initialize POST. The hung probe poisons the function instance, the initialize that follows hangs, and the agent times out after 60 seconds with the server showing "running" and no error logged anywhere. A curl POST smoke test passes throughout, which sends diagnosis in the wrong direction (see #424 for the full wire-level trace).

### Files changed

| File | Change |
|---|---|
| `server/index.ts` | Guard after `app.all("*")` opens, with CORS headers + `Allow` |
| `extensions/_template/AGENT_SPEC.md` | Same guard in the template handler so new extensions don't reintroduce the bug |

### Companion PR

The same `app.all("*")` → `handleRequest` pattern exists in 4 integrations and 2 recipes. A companion PR (`[integrations]`) applies the identical guard there — split out per the PR gate's scoping rules. Extensions using `app.post("*")` are unaffected (Hono 404s other methods before the transport).

### Testing

- Tested against my own live deployment (per CONTRIBUTING: run it yourself first): before the guard, every `mcp-remote` attach timed out at 60s; after deploying it, GET returns 405 in ~1s, POST initialize returns `serverInfo` in ~0.5s, and the full initialize → `tools/list` → tool-call path works through mcp-remote.
- `deno check` output on patched files is identical to `main` (pre-existing import-map resolution notes only; nothing new introduced).

https://claude.ai/code/session_01KpuWP7DYkoWnS2gT8Fd192

### DIFF
--- a/extensions/_template/AGENT_SPEC.md
+++ b/extensions/_template/AGENT_SPEC.md
@@ -148,6 +148,14 @@ const server = new McpServer({
 const app = new Hono();
 
 app.all("*", async (c) => {
+  // Reject non-POST requests: a GET would fall through to
+  // StreamableHTTPTransport.handleRequest and park on an SSE stream that
+  // never emits and never closes, hanging mcp-remote's OAuth-discovery GET
+  // probe and timing out the MCP handshake.
+  if (c.req.method !== "POST") {
+    return c.json({ error: "Method not allowed" }, 405);
+  }
+
   const provided = c.req.header("x-brain-key") || new URL(c.req.url).searchParams.get("key");
   if (!provided || provided !== MCP_ACCESS_KEY) {
     return c.json({ error: "Invalid or missing access key" }, 401);
--- a/server/index.ts
+++ b/server/index.ts
@@ -601,6 +601,17 @@ app.options("*", (c) => {
 });
 
 app.all("*", async (c) => {
+  // Reject non-POST requests up front. This server is stateless over
+  // streamable HTTP: there is no standalone SSE stream (GET) or session
+  // termination (DELETE) to serve. Without this guard a GET falls through to
+  // StreamableHTTPTransport.handleRequest, which parks it on an SSE stream
+  // that never emits and never closes. mcp-remote always sends a GET probe
+  // (OAuth discovery) before its initialize POST, so that probe hangs and
+  // the MCP handshake times out at the client with no error server-side.
+  if (c.req.method !== "POST") {
+    return c.json({ error: "Method not allowed" }, 405, { ...corsHeaders, Allow: "POST, OPTIONS" });
+  }
+
   // Accept access key via header OR URL query parameter
   const provided = c.req.header("x-brain-key") || new URL(c.req.url).searchParams.get("key");
   if (!provided || provided !== MCP_ACCESS_KEY) {


jcpoyser:
Hi @jcpoyser — the branch has been refreshed onto the current main (new head: `4cc184b`).

The underlying hang is still reproducible on main today: `server/index.ts` routes `app.all("*")` directly into `StreamableHTTPTransport.handleRequest` with no method check. A GET probe — which `mcp-remote` always sends for OAuth/WWW-Authenticate discovery before its `initialize` POST — parks on an SSE stream that never emits and never closes; the client times out after 60 s with no server-side error visible. A `curl` POST works fine throughout, which makes the failure easy to misdiagnose.

Is there anything else you'd like addressed before this goes to review?

PR #426, opened 2026-07-10
[integrations] Reject non-POST MCP requests with 405 to prevent GET handshake hang

jcpoyser:
Companion to #425 (core server fix). Addresses the same root cause as #424 in the contribution folders.

### What

Adds the same `405 Method Not Allowed` guard ahead of `StreamableHTTPTransport.handleRequest()` in every integration and recipe that routes `app.all("*")` into the transport. Without it, a GET request parks forever on a never-emitting, never-closing SSE stream inside `@hono/mcp` — and `mcp-remote` always sends a GET probe (OAuth discovery) before its initialize POST, so the MCP handshake times out at 60s with no error server-side. Full wire-level diagnosis in #424.

### Files changed

| File | Change |
|---|---|
| `integrations/delete-thought-mcp/index.ts` | Guard after `app.all("*")` opens, with CORS headers + `Allow` |
| `integrations/update-thought-mcp/index.ts` | Same |
| `integrations/kubernetes-deployment/index.ts` | Same |
| `integrations/enhanced-mcp/index.ts` | Same |
| `recipes/edge-function-cost-optimization/examples/after/index.ts` | Same |
| `recipes/edge-function-cost-optimization/README.md` | Same guard in the code snippet so it matches the example; adds an expected-outcome summary |
| `recipes/work-operating-model-activation/index.ts` | Guard exempts OPTIONS — this handler has no `app.options` route above it, so preflight must keep flowing through unchanged |
| `integrations/kubernetes-deployment/README.md` | Adds numbered step overview and the tool-audit guide link (gate requirements for touched folders) |

Not changed: extensions using `app.post("*")` (meal-planning, professional-crm, job-hunt, home-maintenance, family-calendar, household-knowledge) — Hono already 404s non-POST there, no hang possible.

### One design note for maintainers

The cost-optimization recipe reuses transports across requests via a session map. A maximal reading of the spec could support GET-with-`mcp-session-id` for stream resumption there. I went with POST-only anyway because:
- (a) no OB1 server sends server-initiated messages, so an SSE stream has nothing to deliver, and
- (b) mcp-remote's probe GET carries no session id — under the current code it mints a brand-new session, leaks it into the map, and hangs. [If you'd rather support GET-with-session in that recipe, happy to adjust.]

### Testing

Same verification as #425: on a live deployment, GET returns 405 in ~1s (previously hung until killed), POST initialize returns `serverInfo` in ~0.5s, and the full initialize → `tools/list` → tool-call path works through mcp-remote.

https://claude.ai/code/session_01KpuWP7DYkoWnS2gT8Fd192

### DIFF
--- a/integrations/delete-thought-mcp/index.ts
+++ b/integrations/delete-thought-mcp/index.ts
@@ -128,6 +128,17 @@ const app = new Hono();
 app.options("*", (c) => c.text("ok", 200, corsHeaders));
 
 app.all("*", async (c) => {
+  // Reject non-POST requests up front. This server is stateless over
+  // streamable HTTP: there is no standalone SSE stream (GET) or session
+  // termination (DELETE) to serve. Without this guard a GET falls through to
+  // StreamableHTTPTransport.handleRequest, which parks it on an SSE stream
+  // that never emits and never closes. mcp-remote always sends a GET probe
+  // (OAuth discovery) before its initialize POST, so that probe hangs and
+  // the MCP handshake times out at the client with no error server-side.
+  if (c.req.method !== "POST") {
+    return c.json({ error: "Method not allowed" }, 405, { ...corsHeaders, Allow: "POST, OPTIONS" });
+  }
+
   const provided =
     c.req.header("x-brain-key") || new URL(c.req.url).searchParams.get("key");
   if (!provided || provided !== MCP_ACCESS_KEY) {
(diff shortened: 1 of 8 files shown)


justfinethanku:
Merged — thanks! Turning a silent 60-second hang into an instant 405 makes every contributed MCP server far friendlier to set up with mcp-remote. Good catch on the repeated app.all pattern.

PR #436, opened 2026-07-12
[recipes] content-fingerprint-dedup: add write-time fingerprint trigger

eazene:
## What

Adds an optional **write-time fingerprint trigger** to the `content-fingerprint-dedup` recipe (new "Step 2b" in the README).

## Why

`upsert_thought` only dedups when callers actually use it. Any path that writes a raw `INSERT INTO thoughts` — a webhook handler, an Edge Function, a synthesis job, a one-off script — inserts a row with a **NULL** `content_fingerprint` and silently escapes dedup. Because the unique index is partial (`WHERE content_fingerprint IS NOT NULL`), those NULL rows don't conflict with each other, so duplicates accumulate unnoticed.

I hit this in practice: three scheduled Edge Functions writing via raw `INSERT` were leaking ~7–9 NULL-fingerprint rows/week into a live brain.

## How

A `BEFORE INSERT OR UPDATE` trigger (`set_content_fingerprint`) computes the fingerprint at write time for **any** path, so no caller can leak a NULL.

- Normalization is **identical** to `upsert_thought` (`lower` + `trim` + collapse whitespace → SHA-256), so trigger-computed fingerprints line up exactly with the unique index.
- `upsert_thought` is unaffected — it supplies its own fingerprint, the trigger sees a non-NULL value and leaves it alone, and `ON CONFLICT` behaves as before.
- Recomputes on `UPDATE` only when `content` actually changed (`NEW.content IS DISTINCT FROM OLD.content`), so metadata-only updates stay cheap and never touch the fingerprint.

## Notes

- Documented the one behavior change: with the trigger in place, a raw `INSERT` of already-existing content now raises `23505` instead of quietly duplicating. Callers that should merge (retryable webhooks, idempotent re-imports) should go through `upsert_thought`.
- With the trigger installed, the companion `fingerprint-dedup-backfill` recipe becomes a **one-time** cleanup rather than a scheduled job — new writes can no longer leak a NULL.

Docs-only change to an existing recipe (README + `updated` date in metadata.json). No changes to the core `thoughts` table structure.

### DIFF
--- a/recipes/content-fingerprint-dedup/README.md
+++ b/recipes/content-fingerprint-dedup/README.md
@@ -84,6 +84,46 @@ Now call it from your Edge Function or import script:
 SELECT upsert_thought('My thought content', '{"metadata": {"source": "chatgpt"}}'::jsonb);
 ```
 
+### Step 2b: Guarantee a fingerprint on every write (trigger)
+
+`upsert_thought` only dedups when callers actually use it. Any code path that writes a raw `INSERT INTO thoughts` — a webhook handler, an Edge Function, a synthesis job, a one-off script — will insert a row with a **NULL** `content_fingerprint` and silently escape dedup. Because the unique index is partial (`WHERE content_fingerprint IS NOT NULL`), those NULL rows don't even conflict with each other, so the duplicates pile up unnoticed.
+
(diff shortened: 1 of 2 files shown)


justfinethanku:
Merged — thank you. The partial index left direct INSERTs unfingerprinted, and this write-time trigger closes that dedup loophole using the recipe's exact normalization. Idempotent SQL appreciated.

PR #438, opened 2026-07-12
[recipes] provenance-chains: fix eval.mjs default model (dated id 404s on OpenRouter)

eazene:
## What
`eval.mjs` defaults to `anthropic/claude-3.5-haiku`, which OpenRouter no longer serves — every grading call fails with `404 No endpoints found`, so a stock run scores 0/N and exits 1.

## How
Default switched to `anthropic/claude-haiku-4-5` (2 occurrences: the usage comment and the parseArgs default). `--model` still overrides.

Hit live while scheduling the nightly/weekly eval the recipe README recommends. Same failure class as the typed-edge-classifier fix in #415 — dated OpenRouter model ids rot.

### DIFF
--- a/recipes/provenance-chains/eval.mjs
+++ b/recipes/provenance-chains/eval.mjs
@@ -34,7 +34,7 @@
  *   --ids a,b,c                        grade specific thought IDs only
  *   --force                            re-grade thoughts already scored
  *   --dry-run                          skip PATCH back to DB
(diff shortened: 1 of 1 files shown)


justfinethanku:
Merged — thank you. One-line fix, whole recipe unblocked — the dated model id was 404ing every grading call. If you feel like a quick follow-up, the README config example still shows the old id.

PR #446, opened 2026-07-13
[recipes] repo-learning-coach: load .env via dotenv + fix research frontmatter

eazene:
## What & why

The **repo-learning-coach** recipe documents a `.env` workflow (Step 2: copy `.env.example`, fill in Supabase + OpenRouter credentials) but the app never loaded `.env`. Following the README as written, `npm run sync` and `npm run dev` both fail immediately:

```
Error: Missing required environment variable: SUPABASE_URL
```

There was no `dotenv`, no `--env-file`, and Node does not auto-load `.env`, so the server process (`tsx server/index.ts`) started with an empty environment.

## Changes

- Add `dotenv` as a dependency and `import 'dotenv/config'` as the first import in the two server entry points (`server/index.ts`, `server/sync-content.ts`). Placed first so `.env` is loaded before `server/supabase.ts` reads `process.env` at module-init time. The client half (Vite) already loads `.env` on its own; this only affects the Node server process.
- Quote two `research/*.md` frontmatter `summary:` values that contained unquoted colons (`"...the recipe: a local..."`, `"...plain files: one config file..."`). Unquoted colons are parsed as YAML mappings, which broke `npm run sync` with an "incomplete explicit mapping pair" error.

## Testing

On a fresh `.env` (from `.env.example`) against an existing Open Brain Supabase project:

- `npm run sync` → `Synced 3 lessons and 3 research documents.`
- `npm run dev` → server `GET /api/bootstrap` returns HTTP 200 with the project + 3 lessons; Vite client returns HTTP 200.

No schema or app-logic changes; the core `thoughts` table is untouched.

### DIFF
--- a/recipes/repo-learning-coach/package-lock.json
+++ b/recipes/repo-learning-coach/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.0.0",
       "dependencies": {
         "@supabase/supabase-js": "^2.47.10",
+        "dotenv": "^17.4.2",
         "express": "^5.2.1",
         "gray-matter": "^4.0.3",
         "react": "^19.2.0",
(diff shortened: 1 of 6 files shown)


justfinethanku:
Merged — thanks! Loading .env before the DB import (right ESM order, too) and quoting the research frontmatter makes the quickstart actually work as written.

PR #455, opened 2026-07-16
[recipes] Google Activity import: include Gemini answers from safeHtmlItem

araece:
# [recipes] Google Activity import: include Gemini's answers from `safeHtmlItem`

## What

The Google Activity importer currently reads only `title` and `time` from Takeout's
`MyActivity.json` records — which for Gemini Apps means it imports **what the user asked,
but never what Gemini answered**.

It turns out the answers ARE in the export: most Gemini prompt records carry the response
as HTML in a top-level `safeHtmlItem` field:

```json
{
  "title": "Prompted How does polarization work in sunglasses?",
  "time": "2026-03-02T09:15:00.000Z",
  "safeHtmlItem": [{ "html": "<p>Polarized lenses contain a <strong>filter</strong> ..." }]
}
```

Verified against a real export (July 2026): **1,161 of 1,640** Gemini records included it
(median ~420 chars, p90 ~3.7k, max ~86k). The widespread "Takeout exports prompts only"
belief appears to be outdated.

## Changes

- **`extractGeminiAnswer()` + dependency-free `htmlToText()`** — strips the `safeHtmlItem`
  HTML to plain text (tags removed, entities decoded, `<li>` → `- ` bullets), capped at
  1,000 chars per answer.
- **`groupByDay()`** appends `Gemini's answer: …` beneath the prompt title. Records without
  `safeHtmlItem` — all other categories, and the ~30% of Gemini records lacking it — behave
  exactly as before (the helper is a no-op for them).
- **Summarization prompt**: one added paragraph telling the model to use answers to capture
  what was learned/decided, but to attribute facts that appear only in an answer as
  AI-generated ("Gemini answered/suggested …") rather than established truth.
- **Transcript cap** raised 6k → 12k chars for `Gemini Apps` only, so answers survive the
  truncation (~3.4k tokens ≈ $0.0005/day at gpt-4o-mini rates). Other categories keep 6k.
- **README**: documents the discovery.

## Why it matters

Day summaries built from prompts alone can only say what the user was *asking about*.
With the answers included, they can capture what the user actually *found out* — which is
the part worth remembering in a personal knowledge base.

## Requirements

Unchanged from the existing recipe: a working Open Brain setup, an OpenRouter API key,
Node.js 18+. No new services, dependencies, or secrets (`metadata.json` bumped to 1.1.0,
`gemini` tag and description updated).

## Testing

Tested against my own Open Brain instance:

- `node --check` passes.
- Dry-run (`--dry-run --raw --verbose --categories "Gemini Apps"`) against synthetic
  records: answer extracted and appended, HTML/entities cleaned, answerless record
  untouched, other categories unaffected.
- The same parsing approach is running in production on my instance against a real
  1,640-record export.

## Not in scope (possible follow-up)

Localized exports use different filenames/prefixes (German: `MeineAktivitäten.json`,
titles starting `Eingegebener Prompt: `), which `findMyActivityFiles()`/the filters don't
match — non-English Takeout exports are currently invisible to this recipe. Happy to send
that separately if there's interest.

### DIFF
--- a/recipes/google-activity-import/README.md
+++ b/recipes/google-activity-import/README.md
@@ -165,6 +165,8 @@ The filtering is aggressive by design — most Google activity is noise. The scr
 
 **Stage 2: Day grouping & summarization** — Surviving entries are grouped by date. Each day goes to an LLM (gpt-4o-mini via OpenRouter) with a tuned prompt. The LLM extracts 1-3 standalone thoughts per day, focusing on research patterns, decisions, and interests. Days with only trivial activity get empty summaries.
 
+> **Gemini Apps answers:** contrary to what most guides claim, Takeout does not export prompts only — most Gemini prompt records include Gemini's answer as HTML in a `safeHtmlItem` field. The importer strips that HTML to plain text and appends it beneath the prompt (capped at 1,000 chars per answer), so day summaries reflect what was answered, not just what was asked. Records without `safeHtmlItem` behave exactly as before.
+
 **Stage 3: Ingestion** — Each thought gets a vector embedding (text-embedding-3-small, 1536 dimensions) and is inserted into your `thoughts` table with metadata linking back to the source category and date.
 
 ### Deduplication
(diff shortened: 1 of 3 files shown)


justfinethanku:
Merged — thank you. Remembering the answers instead of just the questions is the whole point of a memory system — good catch on safeHtmlItem, and the size cap keeps the cost tiny.

PR #461, opened 2026-07-23
[recipes] Fix UUID id pagination in thought-enrichment backfills

themacmarketer:
## Contribution Type

- [x] Recipe (`/recipes`)

## What does this do?

Fixes a pagination bug in `recipes/thought-enrichment`: `backfill-sensitivity.mjs` and `backfill-type.mjs` initialize their keyset cursor as the number `0` and query PostgREST with `id=gt.0`. On stock Open Brain installs `thoughts.id` is a UUID, so the first request fails with `22P02 invalid input syntax for type uuid: "0"` and both scripts scan 0 rows. The fix seeds the cursor with the zero UUID (`00000000-0000-0000-0000-000000000000`), which sorts before every real UUID, so keyset pagination works unchanged.

`enrich-thoughts.mjs` is unaffected — it seeds its cursor with `null` and only advances to ids read from fetched rows.

## Requirements

No new requirements — same Node.js 18+ and Supabase setup as the existing recipe.

## Checklist

- [x] My contribution has a `README.md` with prerequisites, step-by-step instructions, and expected outcome (unchanged — bug fix to existing recipe scripts)
- [x] My `metadata.json` has all required fields (unchanged)
- [x] If my contribution depends on a skill or primitive, I declared it in metadata.json and linked it in the README (n/a)
- [x] I tested this on my own Open Brain instance (856-row install: both scripts previously failed on the first request; after the fix they paginate the full table — sensitivity backfill scanned 856 rows and upgraded 5, type backfill processed all `reference` rows)

### DIFF
--- a/recipes/thought-enrichment/backfill-sensitivity.mjs
+++ b/recipes/thought-enrichment/backfill-sensitivity.mjs
@@ -102,7 +102,7 @@ console.log();
 
 const BATCH_SIZE = 500;
 const PROGRESS_EVERY = 5000;
-let afterId = 0;
+let afterId = "00000000-0000-0000-0000-000000000000";
 let scanned = 0;
 let scannedAtLastProgress = 0;
 let upgradedPersonal = 0;
--- a/recipes/thought-enrichment/backfill-type.mjs
+++ b/recipes/thought-enrichment/backfill-type.mjs
@@ -166,11 +166,12 @@ async function main() {
   console.log(`Batch size: ${BATCH_SIZE}`);
   console.log("");
 
-  // Cursor replaces offset. afterId starts at 0 (all thought ids are
-  // positive) and advances to the last id seen in each page, so a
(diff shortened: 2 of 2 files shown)


themacmarketer:
Note on the failing **OB1 Review** check: the failure is in the workflow's checkout step, not this change — `actions/checkout` refuses to fetch fork PR code inside a `pull_request_target` workflow (`allow-unsafe-pr-checkout: false` guard). Every fork-based PR will hit this until the workflow is adjusted (e.g. splitting the untrusted checkout into a plain `pull_request` job, per https://gh.io/securely-using-pull_request_target). The diff itself is a two-line cursor-seed fix with no CI-relevant surface.

justfinethanku:
Merged — thanks! The id=gt.0 seed against a UUID column silently stopped both backfills before they scanned a single thought; this makes them actually run on stock installs. (The red OB1 Review check was our workflow's fork-checkout protection, not your code.)

PR #472, opened 2026-08-01
[integrations] Discord Capture via slash command + message command

whiscard:
## Summary
- Replaces the Discord Capture stub (which had no real steps, just a TODO placeholder) with a working design.
- Discord's Interactions webhook only fires for commands, not passive channel messages — a stateless Supabase Edge Function can't hold a Gateway connection open the way Slack/Telegram capture can. So this implements a `/capture <text>` slash command and a "Capture to Open Brain" message context-menu command, both handled by one Edge Function.
- Handles Ed25519 signature verification, the PING handshake, and Discord's 3-second ACK window via a deferred response + `EdgeRuntime.waitUntil` background processing + follow-up webhook patch, matching the embed/classify/insert pattern used by slack-capture and telegram-capture.
- Dedupes message-command captures by `discord_message_id`; updated `metadata.json` description/tags/`updated` date to match.

## Test plan
- [ ] `supabase functions deploy discord-capture --no-verify-jwt` succeeds
- [ ] Interactions Endpoint URL saves in the Developer Portal (PING/PONG handshake passes)
- [ ] `/capture some text` produces an ephemeral confirmation and a `thoughts` row with `metadata.source = 'discord'`
- [ ] "Capture to Open Brain" on an existing message produces a `thoughts` row and dedupes on repeat use
- [ ] `metadata.json` passes the repo's schema validation gate

### DIFF
--- a/integrations/discord-capture/README.md
+++ b/integrations/discord-capture/README.md
@@ -1,74 +1,456 @@
 # Discord Capture
 
-> Capture messages from Discord channels into Open Brain — the same pattern as Slack capture, but for Discord.
+> **Add Discord as a quick-capture interface for your Open Brain.** Type `/capture` in any channel, or right-click an existing message and choose "Capture to Open Brain" — either way it's embedded, classified, and stored, with an ephemeral confirmation back in Discord.
+
+---
 
 ## What It Does
 
-A Discord bot that monitors designated channels and captures messages into Open Brain as thoughts. Messages are embedded and stored with Discord-specific metadata (server, channel, author, timestamp). Works just like the built-in Slack capture but for Discord communities.
+Runs a Supabase Edge Function as a Discord **Interactions** endpoint. Two capture paths, both handled by the same function:
+
+1. **`/capture <text>` slash command** — type a new thought directly into any channel the bot can see.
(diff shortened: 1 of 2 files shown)


whiscard:
Heads up — the **OB1 Review** check is failing before it even reviews this PR's content:

```
##[error]Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.
```

`.github/workflows/ob1-gate-v2.yml` triggers on `pull_request_target` and its `Checkout PR head safely` step tries to check out `refs/pull/${{ github.event.pull_request.number }}/head` directly. `actions/checkout@v4` blocks that by default for any PR coming from a fork, since `pull_request_target` runs with the base repo's secrets/token — this is GitHub's built-in guard against exactly that combination, and it requires the workflow to explicitly set `allow-unsafe-pr-checkout: true` to opt in.

This isn't specific to this PR's content — `label` and `welcome` both passed fine, and this would block the review gate on any fork-based external contribution, not just this one. Flagging in case it's not already on your radar. Happy to adjust anything about this PR itself once the gate can actually run.

PR #475, opened 2026-08-10
Add Novita as a fourth LLM fallback provider

jax-novita:
## Contribution Type

- [x] Integration (`/integrations`)

## What does this do?

Adds Novita as a fourth LLM fallback provider (alongside OpenRouter, OpenAI, Anthropic) in `integrations/smart-ingest` and `integrations/entity-extraction-worker`. Novita exposes an OpenAI-compatible chat completions endpoint, so the new `callNovita()` follows the same structure as the existing `callOpenAI()`.

## Requirements

- Novita API key (`NOVITA_API_KEY`), optional — only needed if you want Novita as one of your configured LLM providers.

## Checklist

- [x] My contribution has a `README.md` with prerequisites, step-by-step instructions, and expected outcome (updated existing READMEs to mention Novita)
- [x] My `metadata.json` has all required fields (unchanged; this PR extends an existing integration rather than adding a new one)
- [ ] If my contribution depends on a skill or primitive, I declared it in metadata.json and linked it in the README (n/a)
- [x] I tested this on my own Open Brain instance (live smoke test against the Novita chat completions endpoint; `deno check` and `deno lint` pass with no new errors/warnings in both integrations)

### DIFF
--- a/integrations/entity-extraction-worker/README.md
+++ b/integrations/entity-extraction-worker/README.md
@@ -24,7 +24,7 @@ The knowledge graph enables queries like "what projects does Sarah work on?" or 
 - Working Open Brain setup ([guide](../../docs/01-getting-started.md))
 - **Enhanced thoughts schema** applied — install `schemas/enhanced-thoughts`
 - **Knowledge graph schema** applied — install `schemas/knowledge-graph` to create the `entities`, `edges`, `thought_entities`, and `entity_extraction_queue` tables
-- At least one LLM API key: OpenRouter (recommended), OpenAI, or Anthropic
+- At least one LLM API key: OpenRouter (recommended), OpenAI, Anthropic, or Novita
 - Supabase CLI installed for deployment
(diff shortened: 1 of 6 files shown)


jax-novita:
Hi, I checked the failed `OB1 PR Gate` run.

The failure appears to come from the repository workflow rather than from this PR's code. The workflow is triggered by `pull_request_target` and attempts to check out the fork head branch, which GitHub refuses for security reasons:

> Refusing to check out fork pull request code from a `pull_request_target` workflow.

There are also warnings about the Node.js 20 runtime used by `actions/checkout@v4` / `actions/upload-artifact@v4`, and the artifact upload step reports that `gate-artifact/` does not exist after the failed review step.

Could you please take a look at the workflow configuration? I believe the PR should be re-run after the workflow is adjusted. I have not changed the application code since the failure is occurring before the review checks can run.

PR #503, opened 2026-09-15
[recipes] Fix Gmail importer custom-endpoint mode for open-brain-rest /capture

### LINKED ISSUE #502: Bug: Gmail importer endpoint mode broken (wrong header, wrong body shape)
**Which contribution?**
recipes/email-history-import

**What happened?**
When `INGEST_MODE=endpoint` is set, the Gmail importer sends capture requests with the wrong auth header (`x-ingest-key`) and the wrong body shape (`source` + `content_fingerprint` + `extra_metadata`). The open-brain-rest `/capture` gateway authenticates via `x-brain-key` and expects `source_type` + `metadata`. Every request returns 401. The importer never checks `res.ok`, so errors are silently swallowed and the real `thought_id` is never read. Endpoint mode is completely non-functional against any standard open-brain-rest deployment. Direct mode is unaffected.

**Evidence — static diff against the gateway contract**

Current broken code in `recipes/email-history-import/pull-gmail.ts`:
```ts
"x-ingest-key": INGEST_KEY,           // gateway reads x-brain-key
source,                                // gateway expects source_type
content_fingerprint: fingerprint,      // not in captureSchema
if (extraMetadata) body.extra_metadata = extraMetadata;  // not in captureSchema
return (await res.json()) as IngestResult;  // no status check; thought_id never mapped
```

Gateway contract in `supabase/functions/open-brain-rest/index.ts`:
```ts
// auth
c.req.header("x-brain-key")

// captureSchema
content: z.string().min(1),
source_type: z.string().optional(),
metadata: z.record(z.string(), z.unknown()).optional(),

// createThought response
{ thought_id: thoughtId, type, ... }
```

**What did you expect?**
Endpoint mode should send `x-brain-key`, use the `source_type` + `metadata` body shape, surface HTTP errors, and map `thought_id` / `type` from the response into `IngestResult`.

**Steps to reproduce**
1. Configure the Gmail importer with `INGEST_MODE=endpoint`, a valid `INGEST_URL`, and `INGEST_KEY`.
2. Run the importer.
3. Every import silently fails — gateway returns 401, importer logs no error.

**Environment**
- open-brain-rest gateway: current main
- Affects endpoint mode only; direct mode is unaffected

A fix is ready in a PR from @jcpoyser.

jcpoyser:
Fixes #502

## What this does

Corrects `ingestThoughtEndpoint` in `recipes/email-history-import/pull-gmail.ts` so that `INGEST_MODE=endpoint` works against the open-brain-rest `/capture` gateway.

Four problems were present in the function:

| Problem | Was | Fixed to |
|---------|-----|----------|
| Auth header | `x-ingest-key` | `x-brain-key` |
| Body: source field | `source` | `source_type` |
| Body: extra fields | `content_fingerprint` + `extra_metadata` | `metadata: { source, ...extraMetadata }` |
| Response handling | blind cast, no status check | check `res.ok`, map `thought_id` + `type` |

## Gateway contract verification

Verified statically against `supabase/functions/open-brain-rest/index.ts`:

- **Auth**: `auth()` reads `c.req.header("x-brain-key")` — the old `x-ingest-key` was never checked, causing every request to 401.
- **Body**: `captureSchema` accepts `content`, `source_type`, `metadata`. The old `source`, `content_fingerprint`, `extra_metadata` fields are not in the schema.
- **Response**: `createThought()` returns `{ thought_id, type, ... }`. The fix maps those fields into `IngestResult` instead of casting the raw object.

Verification was static only — no live endpoint was used and no real email data was sent anywhere.

## What it requires

- Working Open Brain setup
- Gmail API credentials
- `INGEST_MODE=endpoint` + a running open-brain-rest deployment (only this mode is affected; direct mode is unchanged)

## Scope

Single file changed: `recipes/email-history-import/pull-gmail.ts` — only the `ingestThoughtEndpoint` function. No other importers, no other files.

## Tested

Verified locally that the new body shape and header match the deployed gateway contract. Live endpoint test was not performed (would require real credentials and a running instance).


jcpoyser:
The OB1 Review gate flagged two broken internal links in README.md (lines 103 and 118) that pointed to `../../primitives/content-fingerprint-dedup/README.md`, which no longer exists at that path. The target moved to `recipes/content-fingerprint-dedup/`, so the links now correctly resolve to `../content-fingerprint-dedup/README.md`. The gate validates the whole touched contribution folder, which is why this pre-existing upstream breakage surfaced here. Fixed in the latest commit — no other changes.

